PR #18, opened 2024-09-18
Add fsspec compatible object for interacting with cloud storage

petebachant:
This would be very useful for DVC in order to stop passing data through our server (https://github.com/calkit/calkit-cloud/issues/118), but instead use presigned URLs for private storage, and for public storage, e.g., Zenodo, HF, GitHub releases, etc., we can download from there directly (future work, not implemented here). This also opens up the possibility of connecting Google Drive, Box, etc., to the Calkit Cloud and using OAuth, which saves users from having to define their own OAuth apps for those services.

This should probably be split off into a `calkit-cloud-client` package someday so these features can be used without the rest of Calkit.

### DVC remotes

These can act similar to S3 or GCS buckets, except they require an `owner/project` prefix. It's recommended to keep them bare (no subdirectory) for full compatibility with the Calkit HTTP remote.

```
ck://petebachant/my-project
```

### Plain interaction

```python
fs = calkit.CalkitFileSystem()

with fs.open("ck://petebachant/some-project/files/some-path/to/whatever.png", "rb") as f:
    content = f.read()
```

This will use a presigned URL to read the latest version on the main branch (TODO: future work).

Similarly,

```python
duckdb.sql("SELECT * FROM read_csv('ck://myname/myproject/files/data/*.csv')")
```

should work.

## TODO

- [x] Don't try to set remote auth for `ck://` URLs.
- [x] Make pushing faster by implementing a batched API call for whatever method DVC uses. 
- [x] DVC should indicate that it's making the request via a `for_dvc=True` kwarg in initialization so general storage can go in its own prefix instead of alongside `files/md5`, maybe just `g` for general storage or `non-dvc`. Also, the backend can then be smart enough to fetch the DVC file by its path in the repo, optionally on at some Git ref.
- [x] Fix bug in push where files are not uploaded. This should be tested for sure. May only be a failure on chunked uploads. That's a pain.
- [ ] Handle permanent URLs inside `.calkit/object-urls` (maybe JSON, YAML, or create a directory tree) and look there first. How to handle pushes though? Maybe this data should live in the cloud server, and the `ckfs` object can read/write from whatever URLs the cloud tells it to. Or maybe pushes always go to the Calkit cloud, and when a release is performed, the object URLs are written to the repo.
- [ ] Could this be used more generally for any project-level storage, so should the DVC objects go in a special location in the cloud?
- [ ] Can/should we try to implement the `xet` protocol here?
- [ ] Add `ref` keyword on instantiation and allow fetching the current MD5 corresponding to that file path in DVC based on the Git ref.

## Indexing

```yaml
4d:
  sdfkjhweroifskldhfkjhg35esdf: https://figshare.com/some-file
```

Is this faster than simply using the entire MD5 as the key?

### DIFF
--- a/calkit/environments.py
+++ b/calkit/environments.py
@@ -446,6 +446,8 @@ def check_all_in_pipeline(
         ck_info = calkit.load_calkit_info(wdir=wdir)
     stages = ck_info.get("pipeline", {}).get("stages", {})
     if targets:
+        # Split targets by "@" to handle sub-stages from iterations
+        targets = [t.split("@")[0] for t in targets]
         stages = {k: v for k, v in stages.items() if k in targets}
     envs_in_pipeline = [stage.get("environment") for stage in stages.values()]
     envs_in_pipeline = [


petebachant:
May also want to check out the Object Store Rust library: https://docs.rs/object_store/latest/object_store/

petebachant:
A better approach might be to have `calkit push` make a request to our server to first get temporary credentials and set them as environmental variables for the `dvc push` call. These temporary credentials can have limited access to a bucket prefix, and allow us to interact directly with the cloud storage provider. If we have the endpoint to generate these credentials, these can be used to set these for use with any other software, e.g., DuckDB, that can directly query cloud storage. However, these might need to use `dvcfs`, since data is stored by MD5 rather than its path in the project repo.

PR #84, opened 2024-11-09
Fix issues discovered while running MS Office workflow

petebachant:


calkit/git.py
```diff
@@ -8,7 +8,10 @@
 def detect_project_name(path=None) -> str:
     """Read the project owner and name from the remote.
 
-    TODO: Currently only works with GitHub remotes.
+    TODO: Currently only works with GitHub remotes where the GitHub repo
+    name is identical to the Calkit project name, which is not guaranteed.
+    We should probably look inside ``calkit.yaml`` in ``project.name``
```
petebachant, line 13:
```suggestion
    We should probably look inside ``calkit.yaml`` at ``name``
```

PR #210, opened 2025-01-25
Add remote SSH environment support

### LINKED ISSUE #14: I can define an HPC cluster or other remote (SSH) environment in which to run all or part of my pipeline
What would the path be, an SSH host or something? What would the command be in the workflow file?

Path could be to an SSH env description in `.calkit/envs` or go directly into the env description

## SSH environment config file

```yaml
host: 127.0.0.1
port: 22
user: pete
key: ~/.ssh/my-key
project_dir: ~/my-cluster-folder
```

## Direct definition

```yaml
environments:
  the-cluster:
    kind: ssh
    host: 127.0.0.1
    user: pete
    key: ~/.ssh/my-key
    sync_paths:
      - data
      - scripts
```

## What happens when we run a command in the remote environment?

1. We should probably sync files back and forth. Maybe this belongs in the environment definition itself. Should we use Git or DVC for this, or just `scp`?
2. Should we have Calkit installed there? We probably assume that the environment is managed by some other entity and our local Calkit instance is not responsible for managing environments there.

petebachant:
## TODO

- [x] Allow jobs to continue running even if disconnected, where if we run the same command again, we ensure the previous has exited on the remote machine before copying back any files.
- [ ] Add unit test for this? We could just SSH to localhost.
- [ ] Fails on Windows because Git checks out files with Windows line endings, which get sent to the host as-is.
- [x] Docs for environments and examples page
- [ ] Ability to simply get a shell in that environment -- perhaps `calkit shell` would be better for that
- [ ] Default copy to and back? Also allow setting that in the command?

Resolves #14 

Example project: https://github.com/calkit/example-ssh

### DIFF
--- a/calkit/__init__.py
+++ b/calkit/__init__.py
@@ -1,4 +1,4 @@
-__version__ = "0.16.3"
+__version__ = "0.17.0"
 
 from .core import *
 from . import git
--- a/calkit/cli/main.py
+++ b/calkit/cli/main.py
@@ -3,6 +3,7 @@
 from __future__ import annotations
 
 import csv
+import glob
 import os
 import platform as _platform
 import subprocess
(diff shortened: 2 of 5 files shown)


petebachant:
Something like this run on the remote machine can give us the PID:

```sh
nohup bash script.sh > lastjob.out 2>&1 & echo $! > lastjob.txt
```

PR #262, opened 2025-02-19
Add Calkit pipeline definition

### LINKED ISSUE #218: I want it to be even easier than DVC to define a pipeline
Imagine something like this in `calkit.yaml`:

```yaml
pipeline:
  get-data:
    kind: python-script # Implied from environment and script?
    script: scripts/get-data.py
    environment: main
  plot-weekly-dists:
    kind: python-script
    script: scripts/plot-weekly-dists.py
    environment: main
  plot-weekly-energy:
    kind: python-script
    script: scripts/plot-weekly-energy.py
    environment: main
  build-paper:
    kind: latex
    environment: tex
    target: paper/paper.tex
    deps_from_stage_outs: 
      - plot-weekly-energy
```

We could use an LLM or other analysis to detect inputs and outputs, compile to `dvc.yaml`, then run.

If the pipeline is defined in `calkit.yaml`, we should note in `dvc.yaml` that it was auto-generated, and overwrite when executing `calkit run`.

We could also do things like add `input_from_stage_outs` keys.

Should we have a `stages` subobject so we can define things about the pipeline, e.g., its type? We should make these stages supersets of DVC stage objects, so they can be written very similarly.

Can we auto-detect outputs? Can we auto-detect deps?

petebachant:
Resolves #218 

## TODO

- [ ] Finish design
- [ ] Finish implementation

### DIFF
--- a/calkit/models/pipeline.py
+++ b/calkit/models/pipeline.py
@@ -1,0 +1,35 @@
+"""Pipeline models."""
+
+from __future__ import annotations
+
+from typing import Literal
+
+from pydantic import BaseModel
+
+
+class PipelineStep(BaseModel):
+    """A step in the pipeline, analogous to a stage in DVC syntax."""
+
+    kind: Literal[
+        "dvc",
+        "python-script",
+        "bash-script",
+        "sh-script",
+        "matlab-script",
+        "sh-command",
+        "bash-command",
+    ]
+    environment: str | None = None
+    wdir: str | None = None
+    inputs: list[str]
+    outputs: list[str, dict]
+    always_run: bool = False
+
+
+class Pipeline(BaseModel):
+    """A computational pipeline, which will typically be transpiled to a DVC
+    pipeline.
+    """
+
+    kind: Literal["dvc"] = "dvc"
+    steps: dict[str, PipelineStep] = {}
--- a/calkit/__init__.py
+++ b/calkit/__init__.py
@@ -1,4 +1,4 @@
-__version__ = "0.20.3"
+__version__ = "0.21.0"
 
 from .core import *
 from . import git
--- a/calkit/models/__init__.py
+++ b/calkit/models/__init__.py
@@ -1,0 +1,1 @@
+from .core import *
--- a/calkit/models/core.py
+++ b/calkit/models/core.py
@@ -7,6 +7,8 @@ from typing import Literal
 
 from pydantic import BaseModel
 
+from calkit.models.pipeline import Pipeline
+
 
 class _ImportedFromProject(BaseModel):
     project: str
@@ -259,6 +261,8 @@ class ProjectInfo(BaseModel):
     git_repo_url: str | None = None
     derived_from: DerivedFromProject | None = None
     questions: list[str] = []
+    dependencies: list = []
+    pipeline: Pipeline | None = None
     datasets: list[Dataset] = []
     figures: list[Figure] = []
     publications: list[Publication] = []
--- a/docs/pipeline/index.md
+++ b/docs/pipeline/index.md
@@ -1,9 +1,40 @@
 # The pipeline
 
-The pipeline, implemented with
-[DVC](https://dvc.org) and stored in the `dvc.yaml` file,
+The pipeline
 defines the processes that produce
 the project's important artifacts.
+It can either be written in Calkit's syntax in the `pipeline` object
+of `calkit.yaml` or in
+[DVC](https://dvc.org)'s syntax and stored in the `dvc.yaml` file.
+If written in Calkit's syntax, a `dvc.yaml` file will be created at
+run time and executed with DVC.
+
+## Calkit pipeline syntax
+
+Calkit's pipeline syntax defines `steps`,
+each of which can have `inputs`, `outputs`, and a `kind` attribute
+to indicate what kind of process is run,
+e.g., a Python script, MATLAB script, R script, shell command, etc.
+
+In the `calkit.yaml` file, you can define a `pipeline` like:
+
+```yaml
+pipeline:
+  steps:
+    collect-data:
+      kind: python-script
+      environment: main
+      script: scripts/collect-data.py
+      outputs:
+        - data/raw.csv
+        - path: data/raw.csv
+          type: dataset
+          title: The raw data
+          description: Raw voltage, collected from the sensor.
+          store_with: dvc
+```
+
+## DVC pipeline syntax
 
 The command, or `cmd` key, for each stage should typically
 include the `calkit xenv` command,


calkit/models/pipeline.py
```diff
@@ -0,0 +1,35 @@
...
+        "python-script",
+        "bash-script",
+        "sh-script",
+        "matlab-script",
+        "sh-command",
+        "bash-command",
+    ]
+    environment: str | None = None
+    wdir: str | None = None
+    inputs: list[str]
```
petebachant, line 24: Could be a dict just as well.

calkit/models/pipeline.py
```diff
@@ -0,0 +1,35 @@
...
+    outputs: list[str, dict]
```
petebachant, line 25: Maybe we want a `PipelineOutput` model.

rebeccamccabe:
For the `src/some_m_file.m` input dependency, it might be less error prone to get them with matlab's dependency analysis, so the pipeline definition only worries about data inputs not code inputs.
ie for my project:
```
>> f = matlab.codetools.requiredFilesAndProducts('simulation.m').'

f =

  24×1 cell array

    {'C:\Users\rgm222\Documents\Github\SEA-Lab\MDOcean\mdocean\plots\util\bluewhitered.m'                                          }
    {'C:\Users\rgm222\Documents\Github\SEA-Lab\MDOcean\mdocean\plots\util\hatchfill2.m'                                            }
    {'C:\Users\rgm222\Documents\Github\SEA-Lab\MDOcean\mdocean\plots\util\improvePlot.m'                                           }
    {'C:\Users\rgm222\Documents\Github\SEA-Lab\MDOcean\mdocean\plots\util\imrange.m'                                               }
    {'C:\Users\rgm222\Documents\Github\SEA-Lab\MDOcean\mdocean\plots\util\signed_log.m'                                            }
    {'C:\Users\rgm222\Documents\Github\SEA-Lab\MDOcean\mdocean\simulation\modules\MEEM\group_velocity.m'                           }
    {'C:\Users\rgm222\Documents\Github\SEA-Lab\MDOcean\mdocean\simulation\modules\MEEM\run_MEEM.m'                                 }
    {'C:\Users\rgm222\Documents\Github\SEA-Lab\MDOcean\mdocean\simulation\modules\dynamics\dynamics.m'                             }
    {'C:\Users\rgm222\Documents\Github\SEA-Lab\MDOcean\mdocean\simulation\modules\dynamics\get_abc_symbolic.m'                     }
    {'C:\Users\rgm222\Documents\Github\SEA-Lab\MDOcean\mdocean\simulation\modules\dynamics\get_dynamic_coeffs.m'                   }
    {'C:\Users\rgm222\Documents\Github\SEA-Lab\MDOcean\mdocean\simulation\modules\dynamics\get_hydro_coeffs.m'                     }
    {'C:\Users\rgm222\Documents\Github\SEA-Lab\MDOcean\mdocean\simulation\modules\dynamics\get_hydro_coeffs_MEEM.m'                }
    {'C:\Users\rgm222\Documents\Github\SEA-Lab\MDOcean\mdocean\simulation\modules\dynamics\multibody_impedance.m'                  }
    {'C:\Users\rgm222\Documents\Github\SEA-Lab\MDOcean\mdocean\simulation\modules\dynamics\multibody_response.m'                   }
    {'C:\Users\rgm222\Documents\Github\SEA-Lab\MDOcean\mdocean\simulation\modules\dynamics\pick_which_root.m'                      }
    {'C:\Users\rgm222\Documents\Github\SEA-Lab\MDOcean\mdocean\simulation\modules\dynamics\spar_dynamics.m'                        }
    {'C:\Users\rgm222\Documents\Github\SEA-Lab\MDOcean\mdocean\simulation\modules\econ\LCOE_from_capex_design_power.m'             }
    {'C:\Users\rgm222\Documents\Github\SEA-Lab\MDOcean\mdocean\simulation\modules\econ\econ.m'                                     }
    {'C:\Users\rgm222\Documents\Github\SEA-Lab\MDOcean\mdocean\simulation\modules\geometry.m'                                      }
    {'C:\Users\rgm222\Documents\Github\SEA-Lab\MDOcean\mdocean\simulation\modules\structures\float_plate_stress.m'                 }
    {'C:\Users\rgm222\Documents\Github\SEA-Lab\MDOcean\mdocean\simulation\modules\structures\get_plate_stress.m'                   }
    {'C:\Users\rgm222\Documents\Github\SEA-Lab\MDOcean\mdocean\simulation\modules\structures\get_stiffener_equivalent_properties.m'}
    {'C:\Users\rgm222\Documents\Github\SEA-Lab\MDOcean\mdocean\simulation\modules\structures\structures.m'                         }
    {'C:\Users\rgm222\Documents\Github\SEA-Lab\MDOcean\mdocean\simulation\simulation.m'                                            }
```

I normally run everything through github actions rather than locally, which I have set up to use my lab's server, and I just do small debugging locally. So it might be useful to have a pipeline setting for `runs-on: local or github-actions`, and the latter would use the `setup-matlab` and `run-matlab-command` actions, which already handles the batch syntax and builds of simulink or other stuff that needs to be compiled. But I guess this could be replaced with a github action that runs calkit instead of running calkit locally.
Does the "outputs_from_stage" imply that it keeps the outputs in memory without reading and writing files unless necessary? That is handy.
For where the figures are defined, an auto-detect of everything in the figures folder seems reasonable. It seems odd to me for each figure to be linked to a stage, since some figures are usually diagrams made in illustrator/ppt/etc that aren't related to the pipeline but should still be version controlled. How would multiple figures per script, only some of which go into the publication, be handled?

petebachant:
> But I guess this could be replaced with a github action that runs calkit instead of running calkit locally.

I think that would be the ideal approach, and one I've taken in a little [side project](https://github.com/petebachant/strava-analysis).

> Does the "outputs_from_stage" imply that it keeps the outputs in memory without reading and writing files unless necessary?

My initial thought was that it would save some copy/paste for stages with a long list of outputs.

> For where the figures are defined, an auto-detect of everything in the figures folder seems reasonable. It seems odd to me for each figure to be linked to a stage, since some figures are usually diagrams made in illustrator/ppt/etc that aren't related to the pipeline but should still be version controlled.

It's not a requirement for a figure to be linked to a stage; it's just there so users can trace back where one came from. However, even for something that is created manually, like a diagram, one could argue that there is an input file--the one you'd open in PowerPoint, for example--and there's an output, like a PDF. The conversion of the editable to the "published" diagram could be part of the pipeline, thereby ensuring the PDF is re-exported if the input file changes.

> How would multiple figures per script, only some of which go into the publication, be handled?

The project can have many figures in the `figures` list, and not all of them need to make their way into a publication. For example, I believe [this project](https://calkit.io/meganandd/unh-mhkf1/figures) by @meganandd has many figures created by a single stage, and the project doesn't have any publications declared yet. 

So that's kind of the general workflow: Build your simulation/data processing stages, add figure stages, then publications, where not all figures need to be used, but they and their metadata remain declared so they are shared when the project is shared.

rebeccamccabe:
Are the files on the path? It only searches the path. But I’ve never used
that function in batch mode so maybe it’s just not supported.

-Becca

docs/tutorials/openfoam.md
```diff
@@ -179,15 +179,16 @@ and not rerun simulations when they don't need to be rerun.
...
-    --out "cases/{var}/postProcessing" \
-    "laminar" "k-epsilon" "k-omega"
+    --script-path run.py \
+    --environment foam \
+    --arg "--turbulence-model={var}" \
+    --arg "--overwrite" \
+    --input system \
+    --input constant/transportProperties \
+    --output "cases/{var}/postProcessing" \
+    --iter-arg "{var}" "laminar" "k-epsilon" "k-omega" \
```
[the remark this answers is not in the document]
petebachant: Or maybe we should modify this tutorial to simply write it in `calkit.yaml`.

PR #280, opened 2025-03-15
Always map current user ID when running Docker container

### LINKED ISSUE #278: Docker containers that specify user sometimes don't have permissions to access mounted local volume
If we specify the user with `--user $(id -u):$(id -g)`, this should work for non-root containers at least.

petebachant:
Resolves #278


petebachant:
Getting an error with OpenFOAM images that don't set their user explicitly but this doesn't appear to affect the simulations:

image

PR #317, opened 2025-04-11
Add wrapper for DVC CLI and use `python -m dvc` internally

### LINKED ISSUE #315: Calkit should be able to run when `dvc` is not on the PATH
Useful for users who may install with `pipx` or `uv tool install`. They may have DVC installed anyway, but since it's a Python dependency, we'll have the code available in the virtual environment.

petebachant:
Resolves #315 

## TODO

- [x] Replace all `subprocess` calls to `dvc` with `calkit dvc`.
- [x] Should we use `sys.executable, "-m", "dvc"` internally instead of `"calkit", "dvc"` so we aren't creating subprocesses that create subprocesses when we want to run DVC?
- [x] Update docs to recommend installing with `uv tool install`?

### DIFF
--- a/calkit/templates/core.py
+++ b/calkit/templates/core.py
@@ -1,5 +1,7 @@
 """Core functionality for working with templates."""
 
+from __future__ import annotations
+
 import os
 import shutil
 from typing import Literal
--- a/README.md
+++ b/README.md
@@ -35,9 +35,15 @@ If you want to use [Docker](https://docker.com) containers,
 which is typically a good idea,
 that should also be installed.
 For Python, we recommend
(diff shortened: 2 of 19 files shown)


petebachant:
Calling the DVC module directly is quite a bit faster than calling it through `calkit`, so we should probably do that internally. The wrapper CLI command is still useful in some cases though, externally anyway.

calkit/templates/core.py
```diff
@@ -1,5 +1,7 @@
 """Core functionality for working with templates."""
 
+from __future__ import annotations
```
petebachant, line 3: This may be related to one of the issues in https://github.com/calkit/calkit-assistant/issues/24

Issue #274, opened 2025-03-09
Run tests on Windows in GitHub Actions


petebachant:
Many of the failures I've seen happen on Windows, so this is pretty important.

PR #321, opened 2025-04-12
Add uv Docker layer and enable Windows tests

petebachant:
Figured out a workaround for Windows to deal with the fake TTY issue running subprocess in Pytest, which is to force `stdin=subprocess.PIPE` so `sys.stdin.isatty()` returns False.

Towards #274 -- tests at least run okay locally.


petebachant:
Tests now pass on my Windows laptop, but they are quite a bit slower than Linux/macOS.

.github/workflows/test.yml
```diff
@@ -30,9 +30,7 @@ jobs:
           make test
   windows:
     runs-on: windows-latest
-    timeout-minutes: 15
-    # TODO: Figure out why the tests hang and delete line below
-    if: false
+    timeout-minutes: 30
```
petebachant, line 33:
```suggestion
    timeout-minutes: 45
```

PR #363, opened 2025-06-18
Fix another Windows file path slash issue

rebeccamccabe:


rebeccamccabe:
My import command is 
```
$ calkit overleaf import https://www.overleaf.com/project/68516ba5bc36663bb61711a6 pubs/UMERC-2025-grid-value --title "WEC optimization to maximize grid economic value and avoided emissions" --kind conference-paper --sync-path UMERC_2025_grid_value.tex -s references.bib -s logo.jpg --push-path figures
``` 
and it is the slash in pubs/UMERC-2025-grid-value that is causing issues.

Issue #385, opened 2025-06-30
Overleaf sync discrepancy should create merge conflict rather than fail


rebeccamccabe:
I attempted to sync overleaf and evidently there are some conflicts, where I changed the same lines locally that collaborators changed on overleaf. It errors with a failure to apply the diff, but I would want it to do the usual git merge conflict process where <<<< are added to my files so I can resolve the discrepancy manually.
```
$ calkit overleaf sync pubs/UMERC-2025-grid-value/UMERC_2025_grid_value.pdf
Syncing pubs/UMERC-2025-grid-value/UMERC_2025_grid_value.pdf with Overleaf project ID 68516ba5bc36663bb61711a6
Pulling the latest version from Overleaf
Applying to project repo
error: patch failed: pubs/UMERC-2025-grid-value/UMERC_2025_grid_value.tex:173
error: pubs/UMERC-2025-grid-value/UMERC_2025_grid_value.tex: patch does not apply
error: patch failed: pubs/UMERC-2025-grid-value/references.bib:699
error: pubs/UMERC-2025-grid-value/references.bib: patch does not apply
Error: Failed to apply diff

```
Windows 10, calkit 0.23.1.

PR #386, opened 2025-06-30
Add force option to overleaf sync

rebeccamccabe:
Short term fix to allow manual resolution of overleaf conflicts in response to https://github.com/calkit/calkit/issues/385. This doesn't do an actual merge conflict, but allows the push to happen even if the patch apply fails.


calkit/cli/overleaf.py
```diff
@@ -386,12 +393,15 @@ def sync(
...
+                    ["git", "apply", "--reject", "--directory", wdir, "-"],
                     input=diff,
                     text=True,
                 )
                 if process.returncode != 0:
-                    raise_error("Failed to apply diff")
+                    if force:
+                        typer.echo("Failed to apply Overleaf diff to project repo. Proceeding to push project changes to Overleaf anyway.")
+                    else:
+                        raise_error("Failed to apply Overleaf diff to project repo. Check the .rej files and manually apply changes, then rerun with --force.")
```
petebachant, line 404: I've never used `--reject` before. Where do these `.rej` files end up being saved and what do they look like?

calkit/cli/overleaf.py
```diff
@@ -386,12 +393,15 @@ def sync(
...
                 process = subprocess.run(
-                    ["git", "apply", "--directory", wdir, "-"],
+                    ["git", "apply", "--reject", "--directory", wdir, "-"],
                     input=diff,
                     text=True,
                 )
                 if process.returncode != 0:
-                    raise_error("Failed to apply diff")
+                    if force:
+                        typer.echo("Failed to apply Overleaf diff to project repo. Proceeding to push project changes to Overleaf anyway.")
```
petebachant, line 402: These lines being so long fail the formatting check. You can wrap/continue them like:

```suggestion
                        typer.echo(
                            "Failed to apply Overleaf diff to project repo. "
                            "Proceeding to push project changes to Overleaf anyway."
                        )
```

calkit/cli/overleaf.py
rebeccamccabe: This was my first time using --reject so I'm not an expert. I first tried adding the -3 argument to make it do a 3-way merge, but that failed, probably because the patch is coming from a different repo so it doesn't know how to do the ancestry. So I switched to --reject, which lets it perform the patch where it works and just reject the patches that don't work, rather than erroring without performing any patches. 

The .rej files are in the same folder location as the .tex (or .bib etc) file that threw the error. For me, ` pubs/UMERC-2025-grid-value/UMERC_2025_grid_value.tex.rej`. It looks like a pretty regular diff but not a merge conflict.
```
diff a/pubs/UMERC-2025-grid-value/UMERC_2025_grid_value.tex b/pubs/UMERC-2025-grid-value/UMERC_2025_grid_value.tex	(rejected hunks)
@@ -184,13 +184,15 @@ Temporal and spatial resolution and scope are set to encompass one year each dec
 \subsubsection{Reduced order model and lookup table}
 Linking the design and grid optimization requires first identifying a small number of key design characteristics to collapse the large WEC design space (a reduced order model), and then building a surrogate model to predict the CEM outputs from these reduced design characteristics. The model reduction represents the capture width of the WEC, $CW$, as a fraction of the maximum radiative capture width, $CW_{max}=Gg/\omega^2$. Physics dictates the following form:
 \begin{equation}
+    \label{eq:1}
     \frac{CW}{CW_{max}} = \frac{2 B_p(\omega) \omega^5}{G \rho_w g^3} |RAO(\omega)|^2
 \end{equation}
 where $G$ is the gain (1 for heaving and 2 for surging/pitching), $B_p$ is the generator control damping, $\omega$ is the wave energy frequency, $\rho_w$ is the water density, $g$ is the gravitational acceleration, and $RAO$ is the response amplitude operator (complex ratio between WEC motion on the generator degree of freedom and wave amplitude).
 %In optimal reactive control, $B_p(\omega)=B_h(\omega)+B_d(\omega,RAO(\omega))$, where $B_h$ is the hydrodynamic damping and $B_d$ is nonlinear drag damping. 
-Inspired by a standard second order system with damping ratio $\zeta$ and natural frequency $\omega_n$, we assume the RAO has the form 
+We assume the RAO has the form of a standard second order system with damping ratio $\zeta$ and natural frequency $\omega_n$:
 \begin{equation}
-    RAO(\omega) = \frac{1}{1-(\frac{\omega}{\omega_n})^2 + j 2 \zeta \frac{\omega}{\omega_n}}
+    \label{eq:rao}
+    RAO(\omega) = \frac{1}{1-\left(\frac{\omega}{\omega_n}\right)^2 + j 2 \zeta \frac{\omega}{\omega_n}}
 \end{equation}
 with imaginary unit $j$.
 This implies $B_p = \zeta \omega_n$ if optimal reactive control is used, since the second-order assumption forces $B_p$ to be independent of frequency.
```

petebachant: Gotcha. I wonder if [this package](https://github.com/conan-io/python-patch-ng) would be helpful to apply the rejected patches automatically. Or, it might be nice to walk the user through each rejected hunk and have them specify what should actually go into the document, applying interactively.

rebeccamccabe: From what I can tell, a patch will be rejected if any part of the diff that gets put in the .rej file was changed by both users, even if the changes were on different lines. Ie in my example, I had locally changed some lines above and below the Overleaf changes, but not those lines (I had changed all the B_p's to B_u).

rebeccamccabe: perhaps that package would be useful, although I suspect it would be more worthwhile to provide the necessary ancestor information so a regular merge can take place, and then it can run git mergetool to do a user friendly conflict resolution.

petebachant: Maybe it would be better to use `format-patch` and `am --3way` to merge in every commit since the last sync instead of `diff` and `apply`. I didn't do that initially because of the complexity and the fear that Overleaf would make a huge number of tiny commits automatically, but looking at a project I've been syncing recently, it doesn't seem too bad.

petebachant: A nice thing about `format-patch` is that it will retain commit author information.

rebeccamccabe:
Looking at my overleaf auto commit history, all commits show me as author
although I know my collaborator has made updates, possibly because I am the
overleaf document owner or the one who set up an overleaf git token. But
for the ability to merge, format-patch sounds like a better approach.
Perhaps the overleaf commits could be squashed/rebased when pulled into the
project repo to make the history cleaner.

calkit/cli/overleaf.py
```diff
@@ -386,12 +393,18 @@ def sync(
...
                 )
                 if process.returncode != 0:
-                    raise_error("Failed to apply diff")
+                    if force:
+                        typer.echo(
+                            "Failed to apply Overleaf diff to project repo. "
+                            "Proceeding to push project changes to Overleaf anyway."
+                        )
+                    else:
+                        raise_error("Failed to apply Overleaf diff to project repo. Check the .rej files and manually apply changes, then rerun with --force.")
```
petebachant, line 407:
```suggestion
                        raise_error(
                            "Failed to apply Overleaf diff to project repo. "
                            "Check the .rej files and manually apply changes, "
                            "then rerun with --force."
                        )
```

PR #396, opened 2025-07-02
Use git format-patch and 3-way merge to sync from Overleaf

petebachant:
Related to #386.

### DIFF
--- a/calkit/cli/overleaf.py
+++ b/calkit/cli/overleaf.py
@@ -373,26 +373,39 @@ def sync(
                 "Overleaf config"
             )
         elif last_sync_commit:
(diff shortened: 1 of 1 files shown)


rebeccamccabe:
I gave this a try yesterday with `calkit overleaf sync` on a branch and got some weird behavior probably related to #414 but also relevant to this PR. The branch I was in (`decider-pub`) was behind main in terms of the relevant overleaf paper. Syncing produced two `.rej` files but did not throw the error prompting me to look at them and try again (at least as far as I remember, although my terminal history has cleared so I can't fully confirm. Next time I sync I will see if it happens again). I also have two separate overleaf papers that it was syncing, in case that is relevant.

PR #411, opened 2025-07-10
Only use mamba if installed, solving #410

### LINKED ISSUE #410: Conda env check fails if mamba not installed
I get an error when doing `calkit new conda-env` with conda but not mamba installed.

rebeccamccabe:
Resolves #410

### DIFF
--- a/calkit/conda.py
+++ b/calkit/conda.py
@@ -81,7 +81,7 @@ def check_env(
     info = json.loads(subprocess.check_output(["conda", "info", "--json"]))
     root_prefix = info["root_prefix"]
     envs_dir = os.path.join(root_prefix, "envs")
-    if "mamba" in sys.modules:
+    if calkit.check_dep_exists("mamba"):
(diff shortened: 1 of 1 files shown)


petebachant:
Tests pass, but it looks like the style check is failing. You can ensure commits are checked and fixed automatically if you install the pre-commit hooks with `pre-commit install`.

calkit/conda.py
```diff
@@ -6,6 +6,7 @@
 import os
 import re
 import subprocess
+import sys
```
petebachant, line 9:
```suggestion
```

calkit/conda.py
```diff
@@ -76,12 +76,17 @@ def check_env(
         log_func = calkit.logger.info
     log_func(f"Checking conda env defined in {env_fpath}")
     res = EnvCheckResult()
-    # Use mamba here because it's faster and produces less output
+
```
petebachant, line 79:
```suggestion
```

PR #428, opened 2025-07-18
Allow creating envs from existing spects paths

### LINKED ISSUE #409: I want to use the CLI to add an existing conda env to calkit
It seems like currently in the CLI, `calkit new conda-env` only allows creation of a brand new env, rather than adding an existing `environment.yml` to the `calkit.yml`. In my case, my projects already use conda environments so this would be my preferred interface instead of manually editing the calkit yaml.

petebachant:
Will resolve #409

### DIFF
--- a/calkit/cli/new.py
+++ b/calkit/cli/new.py
@@ -1187,13 +1187,13 @@ def new_conda_env(
 
 @new_app.command("uv-venv")
 def new_uv_venv(
-    packages: Annotated[
-        list[str],
-        typer.Argument(help="Packages to include in the environment."),
-    ],
     name: Annotated[
         str, typer.Option("--name", "-n", help="Environment name.")
     ],
+    packages: Annotated[
+        list[str] | None,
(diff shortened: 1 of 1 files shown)


calkit/cli/new.py
```diff
@@ -1150,21 +1153,34 @@ def new_conda_env(
...
+            repo.git.add(".gitignore")
+        if pip_packages:
+            conda_env["dependencies"].append(dict(pip=pip_packages))
+        with open(path, "w") as f:
+            ryaml.dump(conda_env, f)
+    elif packages is None and os.path.isfile(path):
+        with open(path) as f:
+            conda_env: dict = ryaml.load(f)
+        # Remove prefix
+        conda_env.pop("prefix", None)
```
petebachant, line 1177: @rebeccamccabe curious if this would be a problem for your use case. I'm trying to make the environment more portable by removing any prefix, in case someone exported one with `conda env export`.

PR #467, opened 2025-08-07
Add Julia support

### LINKED ISSUE #182: I want to use a Julia environment and scripts in my pipeline
https://www.reddit.com/r/Julia/comments/164xd0j/how_do_i_manage_virtual_environments_for_julia/

Maybe a good example case: https://github.com/b-fg/HIT.jl

petebachant:
Resolves #182 

## TODO

- [x] Add tests
- [x] Rationalize running script versus including the script -- or does `xenv` run a Julia or shell command?
- [x] Add ability to create a new Julia env from the command line
- [x] Add Julia env check stage to pipeline compilation
- [x] Add docs and example


calkit/cli/check.py
```diff
@@ -544,6 +570,7 @@ def check_matlab_env(
     )
 
 
+@check_app.command(name="julia-env")
```
petebachant, line 573:
```suggestion
```

Issue #496, opened 2025-08-29
I want an easy way to put numerical results from code as text in a latex doc


rebeccamccabe:
I have code that does some calculations and ultimately obtains numerical results, and those numerical results get made into a latex document that defines `\newcommand`s with those values, to be `\include`d in the main latex publication. Script [here](https://github.com/symbiotic-engineering/MDOcean/blob/figures-by-script/mdocean/plots/util/write_vars_to_tex.m), sample latex output [here](https://github.com/symbiotic-engineering/MDOcean/blob/figures-by-script/pubs/journal-paper/numbers.tex) (currently produced manually but would be produced by the script). But I currently don't have an easy way to define different string formatting (ie number of decimal places, presence of a percentage symbol, etc) for each value. 

I see that there is a `calculations` item in `calkit.yaml` and a way to define a template with string formatting for each. This seems like the right sort of structure for what I'm looking for, but I'm not seeing how I could define those calculations as a stage of my pipeline, or mark formatted strings as dependencies of a latex document. The calculations seem more oriented for end-user display, which is not my goal. I'd also like to define each calculation as dependent on a particular pipeline stage, so that if one stage changes only that stage is rerun, as opposed to all of them.

Looking for latex tools to handle the string formatting I found `siunitx` package, and latex tools to import values run from code I found `pythontex` (which supports not just python, seems like I could use matlab via their octave interface), but I'm not clear on the best way to use these with a calkit pipeline.

petebachant:
Any ideas on an ideal interface? I have a project that isn't open source yet that does something similar, writing out a TeX file with a bunch of `\newcommand`s.

One way I could imagine this working is for the MATLAB to export the numerical data in a standard format like CSV or JSON, and then those values could be read in and formatted by Calkit to be included as inputs to the main document:

```yaml
# calkit.yaml
pipeline:
  stages:
    run-calcs:
      kind: matlab-script
      script_path: script.m
      outputs:
        - path: results/vals.json
    build-paper:
      kind: latex
      target_path: paper.tex
      inputs:
        - kind: json-values-to-tex
          input_path: results/vals.json
          output_path: paper/number.tex
          commands: # Must be all lowercase alphanumeric
            myval: {my_val:.1f}\% # Used in TeX document as \myval, referencing my_val key in JSON
            myvalcm: {my_val * 100:.0f} cm
```

With this, you get Python's f-string template syntax so you can do some arithmetic, rounding, add percent symbols or even units.

[...]

PR #503, opened 2025-09-07
Add latexmk command, allow running MATLAB Jupyter notebooks, and add `matlab-command` stage type

petebachant:
## TODO

- [ ] Add option to add to VS Code settings?
- [ ] Add more flexibility?
- [ ] Always prioritize Docker over system environment?
- [ ] Add tests
- [ ] Warn users that they should use the pipeline if one exists, telling them the command for adding the stage
- [ ] Call this `calkit latex build` so we can someday have `calkit latex vals-from-json` for #496?
- [x] Add `matlab-command` stage type

### DIFF
--- a/calkit/cli/notebooks.py
+++ b/calkit/cli/notebooks.py
@@ -169,8 +169,8 @@ def execute_notebook(
             "--language",
             "-l",
             help=(
-                "Notebook language; if 'matlab', will ensure MATLAB kernel is "
-                "available."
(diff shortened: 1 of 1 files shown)


petebachant:
@rebeccamccabe with this implementation all you need to do is add `language: matlab` to your stage definition. Seems like the simplest way to make it work. I was able to run your tutorial notebook after adding that.

Issue #331, opened 2025-05-14
I want to be able to declare my MATLAB environment


petebachant:
Version, toolboxes, path settings (bad)?

The description can be exported with `matlab -batch ver`, and looks something like:

```
WARNING: package sun.awt.X11 not in java.desktop
WARNING: package sun.awt.X11 not in java.desktop
------------------------------------------------------------------------------
MATLAB Version: 24.2.0.2863752 (R2024b) Update 5
MATLAB License Number: 865865
Operating System: macOS  Version: 15.4.1 Build: 24E263 
Java Version: Java 11.0.27+0 with Homebrew OpenJDK 64-Bit Server VM mixed mode
------------------------------------------------------------------------------
MATLAB                                                Version 24.2        (R2024b)
Simulink                                              Version 24.2        (R2024b)
Global Optimization Toolbox                           Version 24.2        (R2024b)
MATLAB Report Generator                               Version 24.2        (R2024b)
Optimization Toolbox                                  Version 24.2        (R2024b)
Parallel Computing Toolbox                            Version 24.2        (R2024b)
Simscape                                              Version 24.2        (R2024b)
Simscape Multibody                                    Version 24.2        (R2024b)
Statistics and Machine Learning Toolbox               Version 24.2        (R2024b)
Symbolic Math Toolbox                                 Version 24.2        (R2024b)
```

rebeccamccabe:
I use `[fList,pList] = matlab.codetools.requiredFilesAndProducts('my_script.m')` to get only the packages used in my project, since I use the same matlab environment for many different projects. This assumes the toolboxes are already installed. Then I manually add them to my list of packages in the setup-matlab github action, because the package naming syntax is frustratingly slightly different between the requiredFilesAndProducts function and [mpm](https://github.com/mathworks-ref-arch/matlab-dockerfile/tree/main/mpm-input-files).

petebachant:
Interesting. I hadn't seen mpm yet. I wonder if the most reproducible approach would be to automate the creation of a Docker image, since MATLAB doesn't have the concept of a virtual environment. I'm sure the GitHub Action does something similar under the hood anyway. How are you ensuring GitHub Actions has access to your license?

rebeccamccabe:
For public repos you don’t actually need a matlab license (which is
surprising). For private repos I think you can configure it somehow in the
CI yaml, I haven’t done that yet.

-Becca

petebachant:
[...]

PR #543, opened 2025-10-08
Matlab auto dependencies

rebeccamccabe:
Feature that uses the matlab `requiredFilesAndProducts` function to find dependencies automatically for matlab stages, as mentioned in https://github.com/calkit/calkit/issues/331. Also adds tests for this functionality. Drawbacks: it is a bit slow, and the `requiredFilesAndProducts` function only works for files, not commands, so for the command stage it creates and then deletes a temporary file `tmp.m`.

### DIFF
--- a/calkit/models/pipeline.py
+++ b/calkit/models/pipeline.py
@@ -339,9 +339,11 @@ class MatlabCommandStage(MatlabStage):
     @property
     def dvc_deps(self) -> list[str]:
         # Write the command to a temporary .m file to get dependencies
-        with tempfile.NamedTemporaryFile() as f:
+        with tempfile.NamedTemporaryFile(mode='w+t', suffix='.m', dir='.') as f:
             f.write(self.command)
         deps = super().get_deps_from_matlab(f.name)
+        print('deps before removing temp file:', deps)
+        print('temp file name:', f.name)
         deps.remove(f.name)
         return deps
 
--- a/calkit/tests/models/test_pipeline.py
+++ b/calkit/tests/models/test_pipeline.py
@@ -173,6 +173,8 @@ def test_matlabcommandstage():
         environment="_system",
         command='disp("Hello, MATLAB!"); matlab_parent',
     )
+    print(s.dvc_deps)
+    print(str(s.dvc_deps))
     assert s.dvc_deps == ["test/matlab_child.m", "test/matlab_parent.m"]
 
 


rebeccamccabe:
This passes tests locally and from my perspective is ready to be merged. 
(`calkit/tests/test_magics.py::test_stage` fails for me because it requires chrome to be installed, but that is unrelated to the functionality).

calkit/tests/models/test_pipeline.py
```diff
@@ -165,6 +166,22 @@ def test_matlabcommandstage():
...
     assert sd["cmd"] == 'matlab -batch "disp(\\"Hello, MATLAB!\\");"'
+    s = MatlabCommandStage(
+        name="d",
+        environment="_system",
+        command='disp("Hello, MATLAB!"); matlab_parent',
+    )
+    assert s.dvc_deps == ["test/matlab_child.m", "test/matlab_parent.m"]
+
+
+def test_matlabscriptstage():
```
petebachant: We can set this to skip if `matlab` isn't available:

```suggestion
@pytest.mark.skipif(shutil.which("matlab") is None)
def test_matlabscriptstage():
```

petebachant (changes requested):
A few minor changes and we should be good to merge. You can also increment the patch version number in `calkit/__init__.py`.

calkit/models/pipeline.py
```diff
@@ -249,13 +252,76 @@ def dvc_outs(self) -> list[str | dict]:
...
+        quoted = filepath.replace("'", "''")
+        # MATLAB code that adds the current folder to path and
+        # prints required files for the given file. Use disp to ensure
+        # newline-separated output.
+        matlab_code = (
+            "addpath(genpath(pwd)); "
+            f"f = matlab.codetools.requiredFilesAndProducts('{quoted}'); "
+            "for i=1:numel(f); disp(f{i}); end"
+        )
+
```
petebachant: There are just a few more blank lines in this method to delete, then we're good to go.

petebachant:
FYI, the MATLAB batch token licensing program is open again: https://www.mathworks.com/support/batch-tokens.html

I am going to request one for the Calkit CI checks, but we should merge this PR once it's ready, with MATLAB tests skipped for now.

petebachant:
I was able to obtain a MATLAB batch token and set it as `MLM_LICENSE_TOKEN` in this repo's Actions secrets, so automated tests should now be possible. Let me know if you want some help getting this over the goal line.

rebeccamccabe:
Yes some help would be appreciated if possible, thanks!

After working with the auto-dependencies in `figures-class-2` branch of MDOcean, I think it is necessary to change the interface of this PR so that it's possible to override the default (current) behavior by specifying a file to use for the auto-dependency that isn't the script/command itself. This is because the matlab internal implementation of auto-dependencies has the behavior that when run for code that uses a class, all methods of that class are listed as dependencies, not just the methods that are called. Therefore, without an override, it would list the analysis code as a dependency to the post-processing stage not just the analysis stage, which ruins the point of separating the analysis and postpro stages. Being able to specify that the auto-dep search should run on ie just the file containing the post processing method would resolve this, which is what I now do manually in [`GenericAnalysis.m`](https://github.com/symbiotic-engineering/MDOcean/blob/85349c9bb2296e53d1eabe894320c0d1432d42a6/mdocean/analysis/GenericAnalysis.m#L65).

Second, there is still the issue where no dependencies are detected for the command stage type. The dependency detection for command stages worked when the temporary file was made manually, and broke when switching to the `tempfile` module. I suspect that matlab isn't recognizing the tempfile as valid matlab code. My previous commit tried to ensure that the tempfile has a `.m` extension and is in the right directory, but that wasn't sufficient.

It would also be helpful to explicitly specify a search path for the dependency search, instead of using the current working directory which is what the current PR does. Better yet, allow specifying a function that adds the proper path as part of the environment (I have [`add_mdocean_path.m`](https://github.com/symbiotic-engineering/MDOcean/blob/figures-class-2/add_mdocean_path.m) which I call as part of every matlab stage). Although this is not as important for CI use (my use case) compared to interactive use where the working directory is likely to change.

Issue #551, opened 2025-10-16
I want to save outputs of successful pipeline stages only


rebeccamccabe:
It would be useful to have a `--save-partial` flag or similar for `calkit run` where it saves the outputs of stages that succeeded but doesn't attempt to save the outputs of stages that fail (since those diffs are deletions from when that stage previously succeeded). 

I am running into this use case when debugging incrementally. Say I have 10 stages with sequential dependency, 1-5 are currently succeeding, 6 is failing, and 7-10 are not run because they depend on 6 which failed. Doing `calkit run --save` or `calkit save -am` will save the outputs of 1-5 and also the output of 6 which a deletion-diff. I would like to instead save the outputs of 1-5 but not the output of 6. I'd like to keep the outdated output of stage 6 around so that I have some input data to test stages 7-10 on, before I have fixed stage 6.

PR #576, opened 2025-11-05
Add run option to save outs from successful stages only

rebeccamccabe:
solves https://github.com/calkit/calkit/issues/551 and includes two new test cases


rebeccamccabe:
The two ci test errors, a permission denied on upload to invenio and a julia not found, do not seem to be related to my changes.

calkit/tests/cli/test_main.py
```diff
@@ -538,3 +542,28 @@ def test_stage_run_info_from_log_content():
...
+
+def test_get_outs_from_successful_stages():
+    # Read the same log used in the other test
+    fpath = os.path.join(
+        os.path.dirname(__file__), "..", "..", "..", "test", "test-log.log"
+    )
+    with open(fpath, "r") as f:
+        content = f.read()
+    stage_run_info = _stage_run_info_from_log_content(content)
+
```
petebachant:
```suggestion
```

calkit/tests/cli/test_main.py
```diff
@@ -538,3 +542,28 @@ def test_stage_run_info_from_log_content():
...
+    # Only the two _check-env-* stages completed per the fixture
+    completed = [
+        s
+        for s, info in stage_run_info.items()
+        if info.get("status") == "completed"
+    ]
+    assert sorted(completed) == sorted(["_check-env-py", "_check-env-tex"])
+
```
petebachant:
```suggestion
```

calkit/tests/test_dvc.py
```diff
@@ -35,3 +36,23 @@ def test_get_remotes(tmp_dir):
...
+    outs:
+      - data/model.pkl
+      - results/output.csv
+      - logs/run.log:
+        - cache: false
+"""
+    dvc_yaml.write_text(content)
+    outs = calkit.dvc.get_stage_outputs("build_model")
+    expected = ["data/model.pkl", "results/output.csv", "logs/run.log"]
+
```
petebachant:
```suggestion
```

calkit/cli/main.py
```diff
@@ -1145,8 +1179,17 @@ def run(
         if save_message is None:
             save_message = "Run pipeline"
         if not quiet:
-            typer.echo("Saving the project after successful run")
+            typer.echo("Saving the whole project after successful run")
         save(save_all=True, message=save_message)
+    if save_successes_after_run:
```
petebachant: I think this will need to be moved up into the `if failed` block, and then raise the error. Also, if `save_log` is `False`, we should probably use either a temp file or an `io.StringIO` object so we don't inadvertently write logs that get committed to the repo.

calkit/cli/main.py
```diff
@@ -963,7 +984,19 @@ def run(
     ] = False,
     save_after_run: Annotated[
         bool,
-        typer.Option("--save", "-S", help="Save the project after running."),
+        typer.Option(
+            "--save",
+            "-S",
+            help="Save the project after running, including outputs of successful and unsuccessful stages.",
```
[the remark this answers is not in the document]
rebeccamccabe: Nevermind, I see, because the error would prevent the code from reaching this block, even though it is not inside the if statement.

rebeccamccabe:
It seems like there are potentially four plausible save behaviors:

|                          | --keep-going                                                | no --keep-going                                                    |
|--------------------------|-------------------------------------------------------------|--------------------------------------------------------------------|
| --save-all               | runs A,B,C; saves A,B(deletion),C, and non-pipeline changes | runs A,B; saves A,B(deletion),C(deletion) and non-pipeline changes |
| --save-any-run           | runs A,B,C; saves A,B(deletion),C                           | runs A,B; saves A,B(deletion)                                      |
| --save-any-successful    | runs A,B,C; saves A,C                                       | runs A,B; saves A                                                  |
| --save-if-all-successful | runs A,B,C; saves none                                      | runs A,B; saves none                                               |

where deletion means that saving a failed stage deletes the outputs that were not successfully generated. This is assuming that a failure means the stage did not succeed in generating its outputs. There is potentially even more possibilities considering that it's possible for a stage to throw an error after it successfully generates some or all of its outputs. My original intention of saving only successful stages was to not delete outputs, maybe this would be better accomplished with a `--no-delete` flag or similar instead of using whether the stage succeeded?

rebeccamccabe:
From the documentation of `dvc repro`, it seems like stage ins and outs are auto added and auto committed when the pipeline is run, so the save options in calkit run would only make sense if `--no-commit` is used. 

To address the overlapping output issue, I believe that the stage name rather than the name out the output files should be the target of `dvc add` and `dvc commit` so I will try that.

PR #620, opened 2025-11-30
Use --force with conda env create

petebachant:
Ran into this issue on Windows. Tried to update Conda but still kept at 4.14.0.


petebachant:
Updating both Python and Conda got me up to a newer version with the `-y` option and no `--force` option.

PR #658, opened 2026-01-05
Add optional matlab_path in MatlabScriptStage

rebeccamccabe:
In a matlab-command stage, it's possible to add items to the matlab path as part of the command, but it's not possible in a script stage, so I added a property for this and updated the dvc run command to add the path using that property.

### DIFF
--- a/calkit/models/pipeline.py
+++ b/calkit/models/pipeline.py
@@ -31,6 +31,32 @@ from calkit.notebooks import (
 )
 
 
+def _check_path_relative_and_child_of_cwd(s: str) -> str:
+    p = Path(s)
+    # Enforce that the path is relative
+    if p.is_absolute():
+        raise ValueError(f"Path must be relative: {p}")
+    # Enforce that the path is a child of the (resolved) CWD
+    cwd = Path.cwd().resolve()
+    # Resolve the path relative to the resolved CWD to get a full path for
+    # comparison
+    absolute_path = p.resolve(strict=False)
+    # Check if the absolute path starts with the resolved CWD, ensuring it's a
+    # child
+    try:
+        absolute_path.relative_to(cwd)
+    except ValueError:
+        raise ValueError(
+            f"Path is not a child of the current working directory: {p}"
+        )
+    return p.as_posix()
+
+
+RelativeChildPathString = Annotated[
+    str, AfterValidator(_check_path_relative_and_child_of_cwd)
+]
+
+
 class StageIteration(BaseModel):
     """A model for the ``iterate_over`` key in a stage definition.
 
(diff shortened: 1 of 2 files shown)


calkit/models/pipeline.py
```diff
@@ -380,6 +380,7 @@ def dvc_outs(self) -> list[str | dict]:
 class MatlabScriptStage(Stage):
     kind: Literal["matlab-script"]
     script_path: str
+    matlab_path: str | None = None
```
petebachant: What do you think about enforcing that the path is a relative child path?

```suggestion
    matlab_path: RelativeChildPath | None = None
```

This would require this function and type defined above:

```python
from pydantic import BaseModel, ValidationError, AfterValidator

def validate_relative_and_child_of_cwd(p: Path) -> Path:
    # 1. Enforce that the path is relative
    if p.is_absolute():
        raise ValueError(f"Path must be relative: {p}")
    # 2. Enforce that the path is a child of the CWD
    cwd = Path.cwd()
    # Resolve the path relative to CWD to get a full path for comparison
    absolute_path = (cwd / p).resolve()
    # Check if the absolute path starts with the CWD, ensuring it's a child
    try:
        absolute_path.relative_to(cwd)
    except ValueError:
        raise ValueError(f"Path is not a child of the current working directory: {p}")
    return p

RelativeChildPath = Annotated[Path, AfterValidator(validate_relative_and_child_of_cwd)]
```

calkit/models/pipeline.py
```diff
@@ -192,9 +193,33 @@ def to_dvc(self) -> dict:
...
+    # Resolve the path relative to CWD to get a full path for comparison
+    absolute_path = (cwd / p).resolve()
+    # Check if the absolute path starts with the CWD, ensuring it's a child
+    try:
+        absolute_path.relative_to(cwd)
+    except ValueError:
+        raise ValueError(
+            f"Path is not a child of the current working directory: {p}"
+        )
+    return str(p)
```
petebachant: Might want to do this just to be safe, considering the backslash issues on Windows:

```suggestion
    return p.as_posix()
```

calkit/tests/models/test_pipeline.py
```diff
@@ -201,6 +201,8 @@ def test_matlabscriptstage():
         sd["cmd"]
         == "matlab -batch \"addpath(genpath('scripts')); run('scripts/my_script.m');\""
     )
+    with pytest.raises(ValidationError): # from pydantic import ValidationError
+        s = MatlabScriptStage(..., matlab_path="/some/abs/path")
```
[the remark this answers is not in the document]
rebeccamccabe: Oops! I clicked commit suggestion without looking closely enough

rebeccamccabe:
The remaining failing testcase is from jupyter notebook magics. I'm not sure why it happens since I didn't change anything related to notebooks, and it didn't happen two commits ago.

petebachant:
> The remaining failing testcase is from jupyter notebook magics. I'm not sure why it happens since I didn't change anything related to notebooks, and it didn't happen two commits ago.

Looks like `pathspec` released 1.0, which is incompatible with DVC: https://github.com/treeverse/dvc/issues/10950

I added `pathspec` as an explicit dep here, so we should be good to go.

PR #722, opened 2026-02-25
Detect title in Overleaf import and allow push-only syncing

### LINKED ISSUE #716: I want to be able to link a folder in my project to a fresh Overleaf project, overwriting what's there
```sh
calkit overleaf link . https://overleaf.com/p/fkjsfhkdsjhf
```

Maybe there should be a confirmation that our files will overwrite the ones on Overleaf for the initial link step. We could also have an option to prioritize the files on Overleaf.

petebachant:
- Auto-detect title
- Better detect build stage
- Allow for pushing from local only when importing and syncing

Resolves #716 

## TODO

- [x] Test title detection
- [x] Add a way to prioritize local files on import, e.g., when we created a blank Overleaf project to "export" a local one into
- [x] Use title detection logic from Calkit Cloud
- [x] Refine logging
- [ ] Test push-only option

### DIFF
--- a/calkit/cli/overleaf.py
+++ b/calkit/cli/overleaf.py
@@ -22,10 +22,13 @@ def _extract_title_from_tex(tex_file_path: str) -> str | None:
     """Extract the title from a LaTeX file."""
     from TexSoup import TexSoup
 
-    with open(tex_file_path) as f:
-        overleaf_target_text = f.read()
-    texsoup = TexSoup(overleaf_target_text)
-    return str(texsoup.title.string) if texsoup.title else None
+    try:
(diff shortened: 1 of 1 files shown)


calkit/cli/overleaf.py
```diff
@@ -203,10 +225,45 @@ def import_publication(
             "Target TeX file path cannot be detected; "
             "please specify with --target"
         )
+        return
+    # Try to extract title from the target LaTeX file if not provided
```
petebachant: This logic needs to be merged with the logic above so we only look for candidate targets in the correct place.

Issue #314, opened 2025-04-10
I want to be able to define certain pipeline stages that should only run on certain machines or types of machines


petebachant:
Imagine I want to run a simulation on an HPC but I don't want to compile the paper there. I still want to execute `calkit run`, but I don't want to attempt the other stages. Maybe we can be smarted about the arguments passed to `dvc repro` depending on the machine we're running on.

PR #752, opened 2026-03-12
Add `current-system` environment kind

petebachant:
This is a kind of environment that will export a lock file based on specified properties of the current system, which will then ensure all pipeline stages that use it will run on a system with the same properties.

Towards #314. To finish out that one, we could have a simple `system` env type, and we enforce certain properties or attributes to match, or we fail the env check.

Maybe this should just be called `system`, and there can be detected and enforced properties?

## TODO

- [ ] Allow `current-system` in env check, including as an outer environment
- [ ] Add tests
- [ ] Allow rounding version numbers of system-level dependencies for the lock file


petebachant:
Now obsolete since we have `system` envs.

PR #823, opened 2026-04-14
Add retries to ckfs

petebachant:
Saw a timeout when pushing to GCS.


calkit/fs.py
```diff
@@ -325,6 +327,22 @@ def _execute_operation(
         # Total payload size for progress reporting.
         payload_size = len(data) if data is not None else (file_size or 0)
```
petebachant:
```suggestion
```

PR #830, opened 2026-04-20
Rationalize slurm env setup command usage and detect interpreter for script inputs

petebachant:
## TODO

- [ ] Make sure this is covered in a test
- [ ] Make consistent with options handling, i.e., do we merge or ignore defaults if any options are set in a stage? We currently have a "use" flag, but it should be named something like "merge". This should probably happen in the `calkit slurm batch` function, not the stage command. This way we don't need environment information to mutate the stage. However, we then need to export a slurm env lock as a DVC dep so we invalidate a stage when the default options change if they are being used. This might then invalidate stages that don't use any defaults.

### DIFF
--- a/calkit/cli/slurm.py
+++ b/calkit/cli/slurm.py
@@ -87,7 +87,8 @@ def run_sbatch(
             "--setup",
             help=(
                 "Shell setup command to run before launching the target "
-                "(repeat for multiple commands)."
+                "(repeat for multiple commands). Will ignore environment's "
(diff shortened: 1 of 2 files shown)


petebachant:
We probably want stages to create commands like:

```sh
calkit slurm batch -e {outer_env} {stage_defined_options} --env-defaults=merge|ignore --command "calkit xenv -e {inner_env} -- my command"
```

PR #844, opened 2026-04-28
Add device login flow

### LINKED ISSUE #188: I don't want to have to generate and paste a token into the CLI
That is, setup and OAuth flow from the CLI so we can do something like `calkit login`.

petebachant:
Resolves #188 

## TODO

- [x] Update cloud config docs
- [x] Run this automatically when checking remote auth?
- [x] Friendlier message when device is already authenticated

### DIFF
--- a/calkit/cloud.py
+++ b/calkit/cloud.py
@@ -3,7 +3,6 @@
 from __future__ import annotations
 
 import time
-import warnings
 from functools import partial
 from typing import Literal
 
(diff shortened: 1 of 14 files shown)


calkit/cloud.py
```diff
@@ -15,6 +17,13 @@
 # A dictionary of tokens keyed by base URL
 _tokens = {}
 
+# Single lock guarding all token-refresh operations to prevent thundering herds
+# (e.g. many concurrent fsspec threads all attempting to refresh at once).
```
petebachant:
```suggestion
# (e.g., many concurrent fsspec threads all attempting to refresh at once).
```

PR #867, opened 2026-05-03
Defer imports to speed up initial CLI load

petebachant:
This cuts down initial import time by around 60% or so. Probably worth the extra complexity for the improved UX.


petebachant:
This shaves about 400 ms off of a `calkit -h` call on my local machine, but only 130 ms off a `calkit status` call. Probably still worth the updates if they don't make the code too difficult to read.

PR #887, opened 2026-05-07
Add PBS scheduler support

### LINKED ISSUE #886: I want to be able to run on PBS schedulers
Basically follow the SLURM UX but adapt for PBS. Create a `pbs` environment and allow running stages there, submitting scripts, commands, etc.

petebachant:
Closes #830, resolves #886 

## TODO

- [x] Merging and replacing all environment-defined options and setup are two different behaviors. Maybe use a literal for these: `"ignore" | "replace" | "merge"`
- [x] Define desired default behavior for environment-defined options and setup. We probably want to replace the whole object if anything is defined. Merging or ignoring are opt-in behaviors.
- [x] Generalize "pbs" and "slurm" into simply "scheduler" for options and setup? Maybe even use a `scheduler` subcommand instead of `slurm` and `pbs`? This might break any existing stages out there, but there aren't many.
- [ ] Only export scheduler env lock files for envs that have default options or default setup. Things like hostname won't affect actual runtime behavior? The changes here will invalidate any stages in the wild that use `slurm` envs. Does hostname changing invalidate previous results? Maybe if we're using different hardware and the results depend on that. We could implement the idea for `system` env kinds where we give a list of important attributes to lock.
- [x] Add more general HPC docs with a suggested workflow, environment and stage definition, etc.
- [x] Break backward compatibility with `slurm` key in stages and only allow `scheduler`?
- [x] Break caching of stages that used to have `calkit slurm batch`, and make the command `calkit sched batch`. This is less important if we're invalidating caches due to the new env lock exporting.
- [x] Drop `sbatch` stage kinds altogether? This will really cause havoc, but only my own projects mostly. Can raise an error and say to move to `shell-script`, or auto-convert.


calkit/pipeline.py
```diff
@@ -506,24 +506,37 @@ def to_dvc(
...
+        # ``calkit check env``), not at compile time, so here we only
+        # compute the path and reference it as a DVC dep.
+        if env.get("kind") in ("slurm", "pbs"):
+            lock_fpath = get_env_lock_fpath(
+                env=env, env_name=env_name, wdir=wdir, as_posix=True
+            )
+        else:
+            lock_fpath = get_env_lock_fpath(
+                env=env, env_name=env_name, as_posix=True, for_dvc=True
+            )
```
petebachant: This change probably doesn't need to be here.

calkit/pipeline.py
```diff
@@ -506,24 +506,37 @@ def to_dvc(
...
         if lock_fpath is None:
             continue
         env_lock_fpaths[env_name] = lock_fpath
     project_params = expand_project_parameters(ck_info.get("parameters", {}))
-    # Set any stage slurm options, which requires environment information
-    pipeline.set_stage_slurm_options(environments=environments)
+    # Validate and initialize scheduler options (SLURM/PBS) for stages
+    # whose outer environment is a job scheduler. Env defaults are applied
+    # at job-submission time, not here.
+    pipeline.set_stage_scheduler_options(environments=environments)
```
petebachant, line 940: We shouldn't need this anymore since we're writing the env lock for scheduler envs

calkit/models/pipeline.py
```diff
@@ -131,13 +131,47 @@ def expand_values(
         return vals
 
 
+EnvDefaultsMode = Literal["ignore", "replace", "merge"]
+
+
 class StageSlurmOptions(BaseModel):
```
petebachant: Stage SLURM and PBS options are identical. They don't necessarily represent the same thing, but we could potentially merge into a single `StageSchedulerOptions` model.

calkit/cli/new.py
```diff
@@ -1491,13 +1491,107 @@ def new_slurm_env(
...
+        env["default_options"] = normalized_default_options  # type: ignore
+    if normalized_default_setup:
+        env["default_setup"] = normalized_default_setup  # type: ignore
+    if description is not None:
+        env["description"] = description
+    envs[name] = env
+    ck_info["environments"] = envs
+    with open("calkit.yaml", "w") as f:
+        ryaml.dump(ck_info, f)
+    from calkit.cli.check import check_environment
```
petebachant: Move to top of function

petebachant:
cc @CarolineCardinale in case you want to request any features and/or check the code.

PR #888, opened 2026-05-09
Add subproject pipeline handling

### LINKED ISSUE #124: I want to be able to define subprojects and be able to navigate between them and the parents

petebachant:
Subprojects must be explicitly declared in the parent project like:

```yaml
subprojects:
  - path: path/to/subproject/folder
```

After that, `calkit status`, `calkit run`, etc., by default will act on all subprojects, allowing DVC to properly compute the DAG and run subproject stages from the parent if applicable.

For https://github.com/calkit/calkit/issues/124#issuecomment-4409076282

I suppose this resolves #124 

## TODO

- [x] It looks like DVC doesn't pick up sub-pipelines in submodules if they have their own `.dvc` folders.
- [ ] Add `--no-subprojects` flag to `run` to disable this?
- [x] Rich output during `status` significantly slowed things down?
- [ ] Raise a warning when trying to get pipeline status in a project with no pipeline?
- [ ] ~~Subprojects should simply be auto-detected?~~ Not for now.
- [x] Not a huge fan of the preceding underscore for subproject wrapper stage names as they show up in the out of `calkit run`.
- [x] Make sure we delete wrapper stages (i.e., they're gone when recompiling DVC pipeline) when subprojects are removed.
- [ ] Subproject could be a stage kind rather than something declared at the project level. This then allows users to control subprojects running in the superproject pipeline definition?


calkit/pipeline.py
```diff
@@ -477,16 +539,32 @@ def to_dvc(
...
+    for subproject in ck_info.get("subprojects", []):
+        if not isinstance(subproject, dict) or not subproject.get("path"):
+            raise ValueError("Subprojects must have a 'path' defined")
+        sp = Path(subproject["path"]).as_posix()
+        if not os.path.isdir(sp):
+            raise NotADirectoryError(f"Subproject path '{sp}' does not exist")
+        to_dvc(wdir=sp, write=write, verbose=verbose)
     if "pipeline" not in ck_info:
-        raise ValueError("No pipeline found in calkit.yaml")
+        return {}
```
petebachant: We should probably at least raise a warning here.

calkit/tests/cli/main/test_subprojects.py
```diff
@@ -0,0 +1,250 @@
+"""Integration tests for subproject support."""
+
+import os
+import subprocess
+
+import calkit
+import calkit.pipeline
+
+
+def _ck(path, data):
```
petebachant:
```suggestion
def write_ck_info(path, data):
```

calkit/tests/cli/main/test_subprojects.py
```diff
@@ -0,0 +1,250 @@
...
+    """Write a calkit.yaml at ``path``."""
+    with open(path, "w") as f:
+        calkit.ryaml.dump(data, f)
+
+
+def _stage(cmd, inputs=None, outputs=None):
```
petebachant:
```suggestion
def make_stage(cmd, inputs=None, outputs=None):
```

calkit/tests/cli/main/test_subprojects.py
```diff
@@ -0,0 +1,250 @@
...
+        s["inputs"] = inputs
+    if outputs:
+        s["outputs"] = outputs
+    return s
+
+
+# ---------------------------------------------------------------------------
+# Inline subproject
+# ---------------------------------------------------------------------------
+
```
petebachant: Don't do this, and add to AGENTS.md instructions to not do this

```suggestion
```

calkit/tests/cli/main/test_subprojects.py
```diff
@@ -0,0 +1,250 @@
...
+    calkit.pipeline.to_dvc(write=True, manage_gitignore=False)
+    subprocess.check_call(["calkit", "run", "sub1:produce"])
+    with open("sub1/a.txt") as f:
+        assert "v2" in f.read()
+
+
+# ---------------------------------------------------------------------------
+# Isolated subproject
+# ---------------------------------------------------------------------------
+
```
petebachant: No also

```suggestion
```

calkit/cli/main/core.py
```diff
@@ -23,6 +23,7 @@
 from typing_extensions import Annotated, Optional
 
 import calkit
+import calkit.git
```
petebachant: We may not need thing since we already lazy-load this module

PR #955, opened 2026-06-13
Fix bug running notebook with IJulia on Windows

petebachant:


calkit/cli/notebooks.py
```diff
@@ -474,7 +502,7 @@ def execute_notebook(
         # Create this environment and write it to file
         envs[res.name] = res.env
         ck_info["environments"] = envs
-        with open("calkit.yaml", "w") as f:
+        with open("calkit.yaml", "w", encoding="utf-8") as f:
```
petebachant: I think we actually have a function for this in core.py now

calkit/cli/notebooks.py
```diff
@@ -503,15 +531,12 @@ def execute_notebook(
...
-        with open(path, "r") as f:
-            nb_json = json.load(f)
-        metadata = nb_json.setdefault("metadata", {})
-        kernelspec = metadata.setdefault("kernelspec", {})
-        kernelspec["name"] = kernel_name
-        kernelspec["display_name"] = display_name
-        kernelspec.setdefault("language", language.lower())
-        with open(path, "w") as f:
-            json.dump(nb_json, f, indent=1)
+        set_notebook_kernelspec(
```
petebachant: This function is used in one place. Why not put all the logic and the comment inline?

calkit/tests/test_notebooks.py
```diff
@@ -7,6 +7,45 @@
 from git.exc import InvalidGitRepositoryError
 
 import calkit
+from calkit.cli.notebooks import set_notebook_kernelspec
+
+
+def test_set_notebook_kernelspec_preserves_unicode(tmp_path):
+    # Regression: on Windows, opening the notebook without encoding="utf-8"
+    # read/wrote it as cp1252, corrupting non-ASCII code (e.g. the Greek letter
```
petebachant:
```suggestion
    # read/wrote it as cp1252, corrupting non-ASCII code (e.g., the Greek letter
```

calkit/notebooks.py
```diff
@@ -245,7 +245,7 @@ def declare_notebook(
     fpath = os.path.join(wdir, "calkit.yaml")
     ck_info["pipeline"] = pipeline_dict
     ck_info["notebooks"] = notebooks
-    with open(fpath, "w") as f:
+    with open(fpath, "w", encoding="utf-8") as f:
```
petebachant: Again, I think we have a function for writing calkit info, which should use this encoding

Issue #843, opened 2026-04-27
I want it to be easy to generate something ready to upload to the arXiv


petebachant:
This could be a pipeline stage type or a type of release that requires manual uploading (a wizard to confirm once it's there, so we can commit). We can also fail if the project is stale.

PR #977, opened 2026-06-19
Add out_dir to LatexStage and modify dvc_outs

rebeccamccabe:
Add optional out_dir attribute and update dvc_outs method to handle output directory. This is required to have the correct dvc_outs when running with an outdir set in latexmkrc.

### DIFF
--- a/calkit/models/pipeline.py
+++ b/calkit/models/pipeline.py
@@ -4,6 +4,7 @@ from __future__ import annotations
 
 import base64
 import json
+import os
 import shlex
 from pathlib import Path
 from typing import Any, Literal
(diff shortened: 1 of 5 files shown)


calkit/models/pipeline.py
```diff
@@ -546,6 +546,7 @@ def dvc_outs(self) -> list[dict]:
 class LatexStage(Stage):
     kind: Literal["latex"] = "latex"
     target_path: str
+    out_dir: str | None = None
```
petebachant: I would probably name this way, with the full word "output", to match `outputs`.

```suggestion
    output_dir: str | None = None
```

petebachant:
I went a bit further with this into checking `latexmkrc` for consistency using the MDOcean project as a guide. Let me know if it's still in line with your intent.

rebeccamccabe:
Yes this still aligns with my workflow, thanks! 

It might also make sense that if there is no latexmkrc found but an out/aux dir is provided, to modify the latexmk command to set outdir and auxdir, so for simple projects a latexmkrc is not needed. The downside here would be that the latexmkrc does not exist for non-Calkit tools like latex workshop to use, so maybe a managed latexmkrc makes more sense, but that seems overly complex. 

I’m also running into frustrations where my latexmkrc is not used when uploading to arxiv, which messes up some of the filepaths that work locally, so maybe keeping Calkit as close as possible to other automatic latex compilers is the best design to avoid this. I’m not sure if the arxiv call is exposed/open source or if there is otherwise a standard setup that is best to mimic. Overleaf accepts latexmkrc but seems like it ignores/overrides/hides the out/aux dirs in a way that I don’t really understand but caused me issues at one point. I am working towards submitting to Elsevier journals and I just read that they require all source files in the same directory with no folders, which means I will have to change my file structure. 

This is now going beyond the scope of this PR but probably useful to keep in mind going forward to ideally reduce the headaches associated with the lack of consistency across different latex compilation workflows.

petebachant:
> I’m also running into frustrations where my latexmkrc is not used when uploading to arxiv, which messes up some of the filepaths that work locally, so maybe keeping Calkit as close as possible to other automatic latex compilers is the best design to avoid this. I’m not sure if the arxiv call is exposed/open source or if there is otherwise a standard setup that is best to mimic. Overleaf accepts latexmkrc but seems like it ignores/overrides/hides the out/aux dirs in a way that I don’t really understand but caused me issues at one point. I am working towards submitting to Elsevier journals and I just read that they require all source files in the same directory with no folders, which means I will have to change my file structure.
> 
> This is now going beyond the scope of this PR but probably useful to keep in mind going forward to ideally reduce the headaches associated with the lack of consistency across different latex compilation workflows.

I've also found the arXiv issue frustrating (#843). There maybe could be additional kinds of stages that transform a LaTeX folder into something flattened and/or ready for arXiv.

PR #980, opened 2026-06-21
Give more control over LaTeX stage behavior

petebachant:


calkit/models/pipeline.py
```diff
@@ -548,11 +548,38 @@ class LatexStage(Stage):
...
     output_dir: str | None = None
+    aux_dir: str | None = None
     latexmkrc_path: str | None = None
     pdf_storage: Literal["git", "dvc"] | None = "dvc"
     verbose: bool = False
     force: bool = False
     synctex: bool = True
+    # Extra arguments passed straight through to latexmk, for control Calkit
+    # does not model. Mirrors the ``args`` field on script stages.
+    args: list[str] = []
```
petebachant: Probably should name this way if we keep it, but it feels a bit too leaky of an abstraction, i.e., `latexmk` could be an internal implementation detail rather than part of the interface in case we want to swap it out for Tectonic, though that would be controlled through the environment. On the other hand, even TinyTeX has a `latexmk`-like interface.

```suggestion
    latexmk_args: list[str] = []
```

rebeccamccabe (approved):
Looks very reasonable! I think the one latexmk feature this doesn’t cover is having multiple targets, which is useful for something like tikz compilation and would require handling list input in the target field of the pipeline, but I currently handle that by parameterization in the pipeline, which would still work with this change.

PR #993, opened 2026-06-28
Fix calkit init to detect existing initialization and require --force

### LINKED ISSUE #391: `calkit init` fails to properly detect existing initialization and fail without `--force`

ChetanyaRathi:
Closes #391

## What changed
`calkit init` now detects when the current directory is already an initialized
Calkit project and exits with a clear error instead of silently re-initializing.
Passing `--force` re-initializes as before.

## Why
Previously, running `calkit init` in an already-initialized project would re-run
and overwrite existing state without warning.

## Behavior
- `calkit init` in an already-initialized project → exits with a non-zero code and
  the message: `This project is already initialized. Use --force to re-initialize.`
- `calkit init --force` → re-initializes as before.
- `calkit init` in a fresh, uninitialized directory → unchanged.

## Testing
- Added a test in `calkit/tests/cli/main/test_core.py` covering the
  already-initialized case (with and without `--force`) and the fresh-directory case.
- `uv run pytest calkit/tests/cli/main/test_core.py::test_init` passes locally.

### DIFF
--- a/calkit/cli/main/core.py
+++ b/calkit/cli/main/core.py
@@ -179,12 +179,22 @@ def init(
         typer.Option(
             "--force",
             "-f",
-            help=("Re-initialize even if the project is already initialized."),
+            help="Re-initialize even if the project is already initialized.",
         ),
     ] = False,
 ):
     """Initialize the current working directory."""
-    if os.path.isfile("calkit.yaml") and not force:
+    import git
+    from git.exc import InvalidGitRepositoryError
+
+    def _project_is_initialized() -> bool:
+        try:
+            git.Repo(".", search_parent_directories=False)
+        except InvalidGitRepositoryError:
+            return False
+        return os.path.isfile(".dvc/config") and os.path.isfile("calkit.yaml")
+
+    if _project_is_initialized() and not force:
         raise_error(
             "This project is already initialized. "
             "Use --force to re-initialize."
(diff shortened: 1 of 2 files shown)


calkit/cli/main/core.py
```diff
@@ -179,11 +179,16 @@ def init(
         typer.Option(
             "--force",
             "-f",
-            help="Force reinitializing DVC if already initialized.",
+            help=("Re-initialize even if the project is already initialized."),
         ),
     ] = False,
 ):
     """Initialize the current working directory."""
+    if os.path.isfile("calkit.yaml") and not force:
```
petebachant: To be considered initialized, a project needs:

1. Git initialized
2. DVC (`.dvc/config` present)
3. `calkit.yaml` present

This only checks for the last one.

calkit/cli/main/core.py
```diff
@@ -179,11 +179,16 @@ def init(
         typer.Option(
             "--force",
             "-f",
-            help="Force reinitializing DVC if already initialized.",
+            help=("Re-initialize even if the project is already initialized."),
```
petebachant: We don't need the parens here as long as we're below 79 characters on this line.

ChetanyaRathi:
Thanks for the review, I updated the PR to check for Git, .dvc/config, and calkit.yaml before treating a project as initialized, and removed the parens from the --force help text. Let me know if anything else needs changing.

Issue #975, opened 2026-06-18
Add a reproducibility check agent skill


petebachant:
This should scan through artifacts looking for magic numbers--things that were copy/pasted but should be referenced as variables. For example, a LaTeX document or Quarto Markdown file with numbers generated by a script, but aren't actually read from the script outputs.

ChetanyaRathi:
Hi @petebachant, I'd like to take #975. I prototyped a deterministic LaTeX checker (it builds provenance from json2latex / \theresults and flags untraceable literals), and the main challenge is false positives — run against the JFM template it currently flags DOIs (10.1017), ORCIDs, citation page ranges, and years, which would be noise in a real paper. So before I finalize, a few questions since it's in Backlog:

Own command calkit check literals, or fold into the existing calkit check repro?
Output: human-readable table, JSON, or both?
Scope: scan all .tex/.qmd, or only publication / latex/quarto pipeline targets in calkit.yaml?
Offline-only deterministic, or is an optional LLM-assisted pass acceptable?
How strict on false positives — should it skip the references/bibliography section entirely, plus years and citation keys?
Should .calkit/ignore.json (from #947) also suppress literal checks for a path?

My instinct for v1: LaTeX-only, skip everything inside \cite/\bibitem/\href/url/doi and the bibliography, ignore years and page ranges, and only flag result-like decimals in prose. Happy to match whatever you prefer.

petebachant:
>Own command calkit check literals, or fold into the existing calkit check repro?

I'd incorporate this into `calkit check repro`.

>Output: human-readable table, JSON, or both?

Both, depending on the `--json` flag, which is useful for the VS Code extension.

>Scope: scan all .tex/.qmd, or only publication / latex/quarto pipeline targets in calkit.yaml?

That's a good question. We'd like to find any targets that are not part of the pipeline, but maybe that's a separate layer of checking.

>Offline-only deterministic, or is an optional LLM-assisted pass acceptable?

I was thinking an agent skill could be okay, but it would call `calkit check repro` to get a deterministic result. Beyond that, an agent could do its own scan for non-reproducible outputs, or those not produced by the pipeline. An LLM-based agent would probably be quite good at that.

>How strict on false positives — should it skip the references/bibliography section entirely, plus years and citation keys?

[...]

PR #994, opened 2026-06-28
Add an artifact ignore file

### LINKED ISSUE #947: I want some way to tell Calkit that a certain artifact is not a certain type or kind so it will ignore it from detection
We're moving towards auto-detecting things like figures, but it's possible there are irrelevant images saved in the project somewhere, and flagging them as figures without provenance would be annoying. We can add something like `.calkit/ignore.json` that tells Calkit to ignore certain paths or to not detect certain paths as certain artifact types.

```json
{
  "path/to/thing": {"not": ["figure", "dataset"]}
}
```

ChetanyaRathi:
Closes #947

## What changed
Adds support for a `.calkit/ignore` file (gitignore-style) that lists paths which
should not be detected as artifacts. During artifact auto-detection, any path
matching a line in `.calkit/ignore` is skipped for all artifact types.

Format (one path or pattern per line, gitignore syntax):

    # example notebook, not a pipeline output
    notebooks/scratch.ipynb
    data/tmp/
    *.bak

Glob and directory patterns are supported. Lines starting with `#` are comments for
human explanation and are not parsed. Blank lines are ignored.

## Why
With auto-detection of figures/datasets, irrelevant files (e.g. a scratch notebook or
a stray image that isn't a real pipeline output) can be flagged as artifacts without
provenance. This gives users a simple, familiar way to mark those paths as
"not an artifact" so they're excluded from detection.

## Behavior
- A path matching a line in `.calkit/ignore` is excluded from all artifact type
  detection (figure, dataset, publication, etc.).
- No `.calkit/ignore` present → detection is unchanged.
- Comments (`#`) let users record why a file doesn't need to be reproducible, kept
  inline in the file rather than as a separate schema.

## Design note
This replaces the earlier `.calkit/ignore.json` per-type schema, per discussion with
@petebachant on this PR. A plain gitignore-ish file matches the `.gitignore` /
`.dvcignore` conventions already in the toolchain (same syntax, glob/directory
patterns for free) and covers the actual use case — a blanket "not an artifact" — more
simply than the per-type JSON. If richer structure is ever needed, the file can be
auto-reformatted later since it lives inside the private `.calkit/` directory.

## Testing
- Reworked `test_detection_ignore` in `calkit/tests/test_detect.py` to cover the plain
  file: listed paths excluded from all types, `#` comments and blank lines ignored,
  glob patterns (`*.ipynb`), directory patterns, and the no-file case. The full
  `test_detect.py` suite (42 tests) passes locally.
- `pre-commit` (ruff, ruff-format, prettier, etc.) passes on the changed files.

### DIFF
--- a/calkit/cli/new.py
+++ b/calkit/cli/new.py
@@ -3174,7 +3174,8 @@ def new_release(
         detected_kind, artifact = calkit.releases.find_artifact(path, ck_info)
         if detected_kind is None:
             detected_kind = calkit.detect.detect_artifact_kind(
-                pathlib.Path(path).as_posix()
+                pathlib.Path(path).as_posix(),
+                ignore=calkit.detect.load_detection_ignore(),
             )
         if release_kind is None:
             if detected_kind is None:
--- a/calkit/cli/new.py
+++ b/calkit/cli/new.py
@@ -3175,7 +3175,6 @@ def new_release(
         if detected_kind is None:
             detected_kind = calkit.detect.detect_artifact_kind(
                 pathlib.Path(path).as_posix(),
-                ignore=calkit.detect.load_detection_ignore(),
             )
         if release_kind is None:
             if detected_kind is None:
--- a/calkit/detect.py
+++ b/calkit/detect.py
@@ -11,7 +11,10 @@ import re
 import shlex
 import sys
 from pathlib import Path
-from typing import Literal
+from typing import TYPE_CHECKING, Literal
+
+if TYPE_CHECKING:
+    import pathspec
 
 NotebookLanguage = Literal["python", "julia", "r"]
 
@@ -1949,41 +1952,20 @@ PRESENTATION_NAMES = {
     "talk.pdf",
 }
 
-DETECTION_IGNORE_FPATH = os.path.join(".calkit", "ignore.json")
+DETECTION_IGNORE_FPATH = os.path.join(".calkit", "ignore")
 
 
-def load_detection_ignore(wdir: str | None = None) -> dict[str, dict]:
-    """Load ``.calkit/ignore.json`` artifact detection exclusions.
+def _load_ignore_spec(wdir: str | None = None) -> "pathspec.PathSpec | None":
+    """Load ``.calkit/ignore`` artifact detection exclusions."""
+    import pathspec
 
-    Returns an empty mapping when the file is absent. Raises ``ValueError`` when
-    the file exists but is not valid JSON.
-    """
     fpath = DETECTION_IGNORE_FPATH
     if wdir is not None:
         fpath = os.path.join(wdir, fpath)
     if not os.path.isfile(fpath):
-        return {}
-    try:
-        with open(fpath, encoding="utf-8") as f:
-            data = json.load(f)
-    except json.JSONDecodeError as e:
-        raise ValueError(f"Invalid JSON in {fpath}: {e}") from e
-    if not isinstance(data, dict):
-        return {}
-    return data
-
-
-def _is_excluded_from_kind(
-    rel_path: str, kind: str, ignore: dict[str, dict]
-) -> bool:
-    """Whether ``rel_path`` must not be auto-detected as ``kind``."""
-    entry = ignore.get(rel_path)
-    if not isinstance(entry, dict):
-        return False
-    not_kinds = entry.get("not")
-    if not isinstance(not_kinds, list):
-        return False
-    return kind in not_kinds
+        return None
+    with open(fpath, encoding="utf-8") as f:
+        return pathspec.PathSpec.from_lines("gitwildmatch", f)
 
 
 def _ancestor_dir_names(rel_path: str) -> set[str]:
@@ -1997,10 +1979,10 @@ def _path_ext(rel_path: str) -> str:
 
 
 def is_figure_path(
-    rel_path: str, *, ignore: dict[str, dict] | None = None
+    rel_path: str, *, ignore: "pathspec.PathSpec | None" = None
 ) -> bool:
     """Whether a repo-relative path looks like an auto-detectable figure."""
-    if ignore and _is_excluded_from_kind(rel_path, "figure", ignore):
+    if ignore is not None and ignore.match_file(rel_path):
         return False
     ancestors = _ancestor_dir_names(rel_path)
     ext = _path_ext(rel_path)
@@ -2014,10 +1996,10 @@ def is_figure_path(
 
 
 def is_dataset_path(
-    rel_path: str, *, ignore: dict[str, dict] | None = None
+    rel_path: str, *, ignore: "pathspec.PathSpec | None" = None
 ) -> bool:
     """Whether a repo-relative path looks like an auto-detectable dataset."""
-    if ignore and _is_excluded_from_kind(rel_path, "dataset", ignore):
+    if ignore is not None and ignore.match_file(rel_path):
         return False
     return (
         _path_ext(rel_path) in DATASET_EXTENSIONS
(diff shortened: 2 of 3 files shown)


petebachant:
There are some merge conflicts now, since this appears to have been based off your branch that worked on the `init` command. Recommend rebasing onto `main`.

ChetanyaRathi:
Rebased onto main and dropped the init commit, that fix is already in via #993, so this is now a single commit with just the .calkit/ignore.json change on top of current main. Conflicts resolved.
Still open from above: do you want a way to ignore a path entirely (e.g. "ignore": true or "not": ["*"]), or keep it strictly per-type for now? Happy to add it either way.

petebachant:
I'm wondering if we need such a complex schema here, i.e., maybe this could be like a gitignore file, `.calkit/ignore`, and any paths in there basically are considered not-artifacts. For example, someone may have a scratch notebook that is more of an example, not something that produces an important output. To not be warned, they can put the notebook path into `.calkit/ignore` rather than `.calkit/ignore.json`. On the other hand, it might be good to have users define a reason why a certain file doesn't need to be reproducible or have a defined import source.

What do you think?

ChetanyaRathi:
Agreed, plain `.calkit/ignore` over the JSON schema. It matches `.gitignore`/`.dvcignore` (same toolchain, same syntax, no learning curve) and gets glob/directory patterns for free — which the exact-path JSON didn't have. The scratch-notebook case is a blanket "not an artifact" anyway, so `{"not": [...]}` was narrower than the issue needs. We'd lose per-type exclusion, but I don't think that's a real need here (a `path:type` line could layer on later without breaking the format).

For reasons, I'd lean on gitignore's own comments:

```
# example notebook, not a pipeline output
notebooks/scratch.ipynb
```

Simple file, still self-documenting. The only reason to go structured (`path: reason`) is if you want `calkit check repro` to print the reason when it skips a path. So the one fork: reason machine-readable in output, or is a comment in the file enough?

Also lines up with #975 — a path in `.calkit/ignore` is basically "don't expect this to be reproducible," exactly the suppression the literal check needs. Happy to rework this to the plain-file approach once you pick the reason behavior.

petebachant:
Yeah, I think a simple gitignore-ish file is fine, and so is relying on comments for explanations. If we need something more complex in the future, we can auto-reformat, since it's a private file (inside `.calkit`).

ChetanyaRathi:
Reworked to plain `.calkit/ignore` (gitignore syntax, comments for explanations) as discussed, updated the PR description to match. Ready for another look.

calkit/tests/test_detect.py
```diff
@@ -1296,3 +1296,54 @@ def test_detect_project_artifacts_includes_results_and_presentations(tmp_dir):
...
+
+
+def test_detection_ignore(tmp_dir):
+    """``.calkit/ignore`` excludes paths from all artifact kinds."""
+    from calkit.detect import (
+        _load_ignore_spec,
+        detect_artifact_kind,
+        detect_figures,
+    )
+
```
petebachant, line 1320: Your agent appears to be ignoring AGENTS.md some more here.

calkit/cli/new.py
```diff
@@ -3174,7 +3174,8 @@ def new_release(
         detected_kind, artifact = calkit.releases.find_artifact(path, ck_info)
         if detected_kind is None:
             detected_kind = calkit.detect.detect_artifact_kind(
-                pathlib.Path(path).as_posix()
+                pathlib.Path(path).as_posix(),
+                ignore=calkit.detect.load_detection_ignore(),
```
petebachant: If a user is explicitly trying to release an artifact, perhaps we don't need to ignore it.

calkit/detect.py
```diff
@@ -1938,8 +1956,12 @@ def _path_ext(rel_path: str) -> str:
     return ("." + name.rsplit(".", 1)[-1].lower()) if "." in name else ""
 
 
-def is_figure_path(rel_path: str) -> bool:
+def is_figure_path(
+    rel_path: str, *, ignore: "pathspec.PathSpec | None" = None
```
petebachant: Is there a reason to use a string literal for the type annotations here versus importing it?

calkit/tests/test_detect.py
ChetanyaRathi: Went back through AGENTS.md — imported pathspec properly and checked mypy on the changed files; no new type errors in detect.py.

calkit/detect.py
ChetanyaRathi: No reason, imported pathspec at the top and switched to the bare pathspec.PathSpec | None.

calkit/cli/new.py
ChetanyaRathi: Agreed, removed the ignore= lookup from the explicit-release path so the ignore file only affects auto-detection.

calkit/tests/test_detect.py
petebachant: I meant the blank lines within functions, which seems to be in the tests.

docs/cli-reference.md
```diff
@@ -67,9 +67,9 @@ calkit init [OPTIONS]
 
 Options:
 
-| Option          | Type    | Required | Default | Description                                      |
-| --------------- | ------- | -------- | ------- | ------------------------------------------------ |
-| `--force`, `-f` | boolean | no       | False   | Force reinitializing DVC if already initialized. |
+| Option          | Type    | Required | Default | Description                                               |
+| --------------- | ------- | -------- | ------- | --------------------------------------------------------- |
+| `--force`, `-f` | boolean | no       | False   | Re-initialize even if the project is already initialized. |
```
petebachant, line 72: Would this go away after a rebase?

ChetanyaRathi: Good catch, that's a stale doc change from the init work that's already in main via #993, not part of this PR. I'll rebase onto main and drop it so this only contains the artifact-ignore changes.

ChetanyaRathi: Done, rebased and pushed (9bb3379). This diff is gone; PR is now just the artifact-ignore changes.

calkit/tests/test_detect.py
```diff
@@ -1309,3 +1309,53 @@ def test_detect_project_artifacts_includes_results_and_presentations(tmp_dir):
...
+    from calkit.detect import (
+        detect_artifact_kind,
+        detect_figures,
+        load_detection_ignore,
+    )
+
+    # (e) No ignore file -> unchanged detection.
+    assert load_detection_ignore() is None
+    assert detect_figures(["figures/a.png"]) == ["figures/a.png"]
+
```
petebachant:
```suggestion
```

calkit/tests/test_detect.py
```diff
@@ -1309,3 +1309,53 @@ def test_detect_project_artifacts_includes_results_and_presentations(tmp_dir):
...
+                    "# this is a comment",
+                    "",
+                    "paper/figs/report.pdf",
+                    "*.ipynb",
+                    "*.csv",
+                    "notebooks/scratch/",
+                ]
+            )
+        )
+
```
petebachant:
```suggestion
```

calkit/tests/test_detect.py
```diff
@@ -1309,3 +1309,53 @@ def test_detect_project_artifacts_includes_results_and_presentations(tmp_dir):
...
+    ignore = load_detection_ignore()
+
```
petebachant:
```suggestion
```

calkit/tests/test_detect.py
```diff
@@ -1309,3 +1309,53 @@ def test_detect_project_artifacts_includes_results_and_presentations(tmp_dir):
...
+    # (a) a path listed in .calkit/ignore is not detected as ANY artifact type
+    path = "paper/figs/report.pdf"
+    assert detect_artifact_kind(path) == "figure"
+    assert detect_artifact_kind(path, ignore=ignore) is None
+    assert path not in detect_figures([path], ignore=ignore)
+
```
petebachant:
```suggestion
```

calkit/tests/test_detect.py
```diff
@@ -1309,3 +1309,53 @@ def test_detect_project_artifacts_includes_results_and_presentations(tmp_dir):
...
+    # (b) comments and blank lines are ignored (implied by parsing succeeding)
+    assert ignore is not None
+    assert not ignore.match_file("# this is a comment")
+
```
petebachant:
```suggestion
```

calkit/tests/test_detect.py
```diff
@@ -1309,3 +1309,53 @@ def test_detect_project_artifacts_includes_results_and_presentations(tmp_dir):
...
+    # (c) a glob pattern excludes matching files
+    globbed = "data/raw.csv"
+    assert detect_artifact_kind(globbed) == "dataset"
+    assert detect_artifact_kind(globbed, ignore=ignore) is None
+    assert ignore.match_file("analysis.ipynb") is True
+
```
petebachant:
```suggestion
```

petebachant (changes requested):

PR #997, opened 2026-07-01
Add dvc[all] extra

### LINKED ISSUE #992: Inclusion of DVC optional dependencies
Would it be possible to modify `pyproject.toml` to include `dvc[all]`? This would install the various DVC storage backends through uv. This way, if someone is using a custom DVC remote with a different storage backend (e.g. ssh), the storage plug-ins will already be available.

TomJohnZ:
Add dvc[all] extra enabling users to access DVC remotes from all available storage backends.

Alternatively, if you'd prefer to not increase the install size, I could instead add a line to the [Using DVC remotes other than calkit.io](https://docs.calkit.org/cloud-integration/#using-dvc-remotes-other-than-calkitio) docs about installing with `uv tool install --with dvc[all] calkit-python` so that users are aware of the option if they need it; I just realized this was a possibility as I was putting the PR together.

Closes #992

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -30,7 +30,7 @@ dependencies = [
   "bibtexparser",
   "checksumdir",
   "docx2pdf",
-  "dvc==3.67.1",
+  "dvc[all]==3.67.1",
   "fastapi",
   "gitpython",
   "keyring",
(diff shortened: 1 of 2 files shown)


petebachant:
I think this is reasonable. If anyone complains that the install is too large, then we could think about ways to reduce the number of dependencies.

PR #1002, opened 2026-07-02
Map import names to distribution names in env spec (fixes #999)

### LINKED ISSUE #999: When `xr` detects scikit-learn as a dependency via a `sklearn` import, it should not use `sklearn` in the env spec
It should use `scikit-learn`.

ChetanyaRathi:
## Summary

Fixes #999.

`xr`-based detection was writing the **import name** into the generated environment spec rather than the **distribution (package) name**. A file importing `sklearn` produced `sklearn` in the spec, which isn't installable — the PyPI/conda package is `scikit-learn`.

## Reproduction (on `main`)

A script or notebook containing `import sklearn` (or `from sklearn... import ...`) yielded `sklearn` in the detected dependencies instead of `scikit-learn`.

## Change

- Added a module-level `_IMPORT_TO_DISTRIBUTION` mapping in `calkit/detect.py` for the common cases where the import name differs from the distribution name (`sklearn` → `scikit-learn`, plus `cv2`, `PIL`, `bs4`, `yaml`, and a few others).
- Applied it via `.get(top_level, top_level)` at both detection sites (script and notebook paths), so unmapped imports pass through unchanged.
- The mapping is a static table, so detection does **not** require the package to be installed (`requires-python` is `>=3.10`, but detection runs on source imports).

## Tests

- Added a regression test asserting the mapping produces distribution names (covers `sklearn` → `scikit-learn` and a second case to confirm it generalizes).
- Updated two existing tests (`test_detect_python_dependencies_from_script`, `test_detect_dependencies_from_python_notebook`) that asserted the old `sklearn` output; the fixture imports stay as `from sklearn...` since that's the import name in user code — only the detected dependency name changes.

Full `test_detect.py` passes locally (42 tests), ruff + ruff-format clean.

### DIFF
--- a/calkit/detect.py
+++ b/calkit/detect.py
@@ -1549,7 +1549,7 @@ def _is_stdlib_module(module_name: str) -> bool:
     return base_module in common_stdlib
 
 
-_IMPORT_TO_DISTRIBUTION = {
+_PYTHON_IMPORT_TO_DISTRIBUTION = {
     "sklearn": "scikit-learn",
     "skimage": "scikit-image",
     "cv2": "opencv-python",
@@ -1629,7 +1629,9 @@ def detect_python_dependencies(
                     # Get the top-level package name
                     top_level = module_name.split(".")[0]
                     dependencies.add(
-                        _IMPORT_TO_DISTRIBUTION.get(top_level, top_level)
+                        _PYTHON_IMPORT_TO_DISTRIBUTION.get(
+                            top_level, top_level
+                        )
                     )
         elif isinstance(node, ast.ImportFrom):
             if node.module and node.level == 0:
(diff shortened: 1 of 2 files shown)


calkit/detect.py
```diff
@@ -1549,6 +1549,20 @@ def _is_stdlib_module(module_name: str) -> bool:
     return base_module in common_stdlib
 
 
+_IMPORT_TO_DISTRIBUTION = {
```
petebachant: Since this module contains code for multiple languages, we should name this more specifically for now:

```suggestion
_PYTHON_IMPORT_TO_DISTRIBUTION = {
```

ChetanyaRathi:
Good call just renamed to _PYTHON_IMPORT_TO_DISTRIBUTION across the definition and both usage sites.

PR #1003, opened 2026-07-02
Add literal check to calkit check repro (#975)

### LINKED ISSUE #975: Add a reproducibility check agent skill

ChetanyaRathi:
## Add reproducibility literal check to `calkit check repro` (#975)

Closes #975.

Scans manuscript LaTeX for **hardcoded result literals** — result-like numbers typed directly into the document instead of being read from a pipeline output. These are a common reproducibility failure: the number silently goes stale when the analysis changes. The reproducible pattern is to compute the value, emit it to JSON, convert it with `calkit latex from-json`, and reference the generated macro; this check flags values that skip that path.

- Folded into **`calkit check repro`** — no separate command, per the discussion.
- Human-readable table by default; machine-readable output with **`--json`** (for the VS Code extension).
- Flags result-like decimals in prose/math (decimals, scientific notation, values with uncertainty, computed percentages) that aren't traceable to a `from-json` output.
- Each finding carries a self-contained fix suggestion pointing at the `from-json` mapping.
- Deterministic and **static** — reads files, never executes the pipeline; no Docker/DVC/network needed for the check.

### Bias to false negatives (per @petebachant)
Tuned to under-flag rather than be noisy. It skips citations/`\cite`/`\bibitem`, `\href`/`\url`/DOIs, the bibliography, years, and page ranges — and also layout/reference macros (`\includegraphics`, `\setlength`, `\geometry`, `\ref`/`\eqref`) so widths, lengths, and cross-references don't get flagged. Values already traceable to a `from-json` output are suppressed. Integers are not flagged by default.

### Agent skill
Adds a reproducibility-check agent skill under `agent-plugin/skills/` that runs `calkit check repro --json` for the deterministic result first, then can do its own LLM pass for non-reproducible / non-pipeline outputs.

### Deferred (not in this PR)
- `--fix` / auto-creating the `from-json` stage.
- Scanning for pipeline targets that aren't part of the pipeline (the separate checking layer discussed).
- `.calkit/ignore.json` suppression — holding until the #947 schema settles.
- Quarto/`.qmd` support; the deterministic core is LaTeX-only for v1.

### Tests
- `test_check_literals.py` — 21 parametrized cases: each flag type and each exclusion (including the layout-macro cases) as an isolated exact-match assertion, plus a traceable-value case and a finding-schema check.
- `test_check_literals_cli.py` — `check repro` integration exercising the flag path and `--json`.
- Existing `test_check_reproducibility` still passes (no regression to `check repro`).

Opening as a draft for early feedback on scope and the exclusion set before I polish.


petebachant:
I think this is on the right track and could certainly be one of the higher impact features.

ChetanyaRathi:
Rebased onto main now that #994 is merged, so this is just the literal-check changes (check.py, check_literals.py, cli/check.py, the skill, and tests). Marking ready for review and also 22 tests are passing locally.

petebachant:
**From Claude** (Anthropic's Claude Code), prompted by @petebachant. Merged
`main` in and pushed to this branch; it's mergeable again. What I changed and
why:

### The merge

Two conflicts. `calkit/check.py` was **renamed to `calkit/reproducibility.py`**
on main (#1565) and expanded, so this branch's edits to it had to be replayed
there rather than merged. The `calkit/cli/check.py` conflict was entirely main's
Docker-image rewrite (#1578) colliding with the older copy on this branch;
resolved to main's version, since nothing here touches that code.

### Two things that were wrong against main, independent of the merge

**The skill told the agent to hand-write a raw DVC block.** `kind: json-to-latex`
has been a Calkit stage kind for a while, so `cmd: calkit latex from-json` with
hand-written `deps`/`outs` bypasses `calkit.yaml` as the source of truth. The
detection code is fine --- it reads the *compiled* pipeline, so it matches a
`json-to-latex` stage's command correctly --- it was only the guidance that was
stale.

**The skill told the agent to reference `\resultsDragCoefficient`.**
`calkit latex from-json` doesn't generate one macro per key; it generates a
single *keyed* command, so the reference is `\results[DragCoefficient]`. That
macro name would never have existed.

Same root cause inside the checker: the `\newcommand\{\\macro\}\{value\}` regex
in `find_untraceable_literals`'s caller could never match what `from-json`
actually writes, so `macro_name` was always `None` and the `Consider using
\macro` half of every suggestion was dead code. Traceability was in fact coming
from the source JSON values, which is the right place to read it from --- the
results file is what the value *is*, the generated commands are one rendering of
it. I dropped the dead mapping and simplified `from_json_values` to the set form
the signature already accepted.

### Restructuring

The ~100 lines that were inline at the end of `check_reproducibility` are now
four named functions in `reproducibility.py` --- `_traceable_values`,
`_generated_tex_paths`, `_tex_sources`, and
`find_untraceable_literals_in_project` --- so the check reads as a step rather
than a wall, and the parts are reachable from a test. Behavior is unchanged
apart from also handling `foreach` stages, which the inline version missed.

### Tests

`test_check_literals_cli.py` mocked the raw-DVC pattern and a
`\newcommand{\resultsDragCoefficient}{0.42}` file that `from-json` never emits.
It passed, but for the wrong reason, and it documented the pattern the skill got
wrong. Rewritten to declare a real `json-to-latex` stage in `calkit.yaml`,
compile it, and assert against a manuscript that cites one value through
`\result[...]`, types the same value by hand, and types one with nothing behind
it. Plus a second case covering that the file a `json-to-latex` stage *writes*
isn't scanned --- it's full of the very numbers the check looks for, and
reporting them would flag the fix itself.

The 21 parametrized cases in `test_check_literals.py` are untouched and pass.

### Hub

`ReproCheck` is a response model on the hub (`GET .../repro`), so
`untraceable_literals` changes the OpenAPI schema. Regenerated the frontend
client against the running backend; the diff is additive.

### Verified

- `calkit` check tests: 39 passed (literals, CLI, `check` suite)
- reproducibility + schema: 28 passed, 1 skipped
- hub backend, full suite through `tests-start.sh`: 296 passed, 4 skipped
- hub frontend: 154 passed, `tsc --noEmit` clean
- ruff, prettier, biome clean

### Still deferred, as your description says

`--fix` / auto-creating the stage, scanning for pipeline targets outside the
pipeline, `.calkit/ignore.json` suppression (waiting on #947), and `.qmd`
support.

One note for when this lands alongside #1600: that branch adds an
`answer-todos` skill for building the analysis a manuscript's TODOs ask for,
which is the same rule from the construction side. I've pointed the two at each
other --- `answer-todos` now runs `calkit check repro` as its verification step
rather than restating the doctrine, and this skill's "Related" section points
back.

calkit/cli/check.py
```diff
@@ -307,12 +307,51 @@ def check_repro(
...
+                show_header=True,
+                header_style="bold",
+            )
+            table.add_column("File", style="cyan")
+            table.add_column("Line", justify="right")
+            table.add_column("Col", justify="right")
+            table.add_column("Value", style="red")
+            table.add_column("Context")
+            table.add_column("Suggestion")
+
```
petebachant:
```suggestion
```

PR #1004, opened 2026-07-02
Normalize trailing slash/whitespace in overleaf sync path (fixes #985)

### LINKED ISSUE #985: Overleaf folder name is sensitive to trailing spaces
In MDOcean, `calkit overleaf sync pubs/applied-ocean-research` works but `calkit overleaf sync pubs/applied-ocean-research/` (what is populated on tab complete) gives an error that there is no overleaf doc found with that name.

ChetanyaRathi:
## Summary

Fixes #985.

`calkit overleaf sync pubs/applied-ocean-research` works, but the same path with a trailing slash — `pubs/applied-ocean-research/`, which is what shell tab-completion inserts — errored with "no Overleaf doc found." The trailing separator was carried into the doc lookup, so it failed to match the stored key `pubs/applied-ocean-research`.

## Reproduction
calkit overleaf sync pubs/applied-ocean-research     # works
calkit overleaf sync pubs/applied-ocean-research/    # errors: no doc found

## Change

Normalize the input path before the lookup — strip trailing path separators (`/` and `\`) and surrounding whitespace — applied at both the `sync` and `get_status` entry points in `calkit/cli/overleaf.py`. Stored names are untouched; only the user-supplied input is normalized.

The issue title says "trailing spaces" while the body reports a trailing slash, so the fix handles both.

## Tests

Added a regression test that imports and syncs a project using the trailing-slash path form and confirms it resolves correctly (previously errored). Existing overleaf tests unaffected.

### DIFF
--- a/calkit/tests/cli/test_overleaf.py
+++ b/calkit/tests/cli/test_overleaf.py
@@ -362,3 +362,50 @@ def test_extract_title_from_tex(tmp_dir):
         f.write(tex)
     title = _extract_title_from_tex("test.tex")
     assert title == "My Cool Paper"
+
+
+def test_overleaf_sync_trailing_slash_or_space(tmp_dir):
+    pid = str(uuid.uuid4())
+    ol_repo = _make_temp_overleaf_project(pid)
+    ol_repo.git.config(["receive.denyCurrentBranch", "ignore"])
+    subprocess.run(["calkit", "init"], check=True)
+    repo = git.Repo()
+    tmp_remote = (
+        Path(tempfile.gettempdir()) / "overleaf-sync-remotes" / pid
+    ).as_posix()
+    os.makedirs(tmp_remote, exist_ok=True)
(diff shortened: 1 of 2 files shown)


calkit/tests/cli/test_overleaf.py
```diff
@@ -362,3 +362,54 @@ def test_extract_title_from_tex(tmp_dir):
         f.write(tex)
     title = _extract_title_from_tex("test.tex")
     assert title == "My Cool Paper"
+
+
+def test_overleaf_sync_trailing_slash_or_space(tmp_dir):
```
petebachant, line 367: To follow style conventions, we should drop the blank lines in this function

ChetanyaRathi:
Done, dropped the blank lines inside the function.

petebachant:
Thanks @ChetanyaRathi!

PR #1009, opened 2026-07-05
Add sync command group with git, dvc, overleaf, and all targets

### LINKED ISSUE #1007: `sync` should be a subcommand with various targets like `git`, `dvc`, `overleaf`, `all`
To make the CLI more English-like, `calkit sync overleaf` should be an option in addition to `calkit overleaf sync`, for example. We will have more in the future, like `calkit sync gdrive` for syncing any files attached to Google Drive. We'll have Zotero syncing, etc. `calkit sync all` will sync all files from all disparate systems.

ChetanyaRathi:
Closes #1007

Adds a top-level `sync` command group so syncing reads more naturally, with
per-target subcommands:

- `calkit sync git` — pull then push the Git repository
- `calkit sync dvc` — pull then push the DVC repository
- `calkit sync overleaf` — sync folders with Overleaf
- `calkit sync all` — sync every configured system, skipping any that aren't set up

## Backward compatibility

`calkit overleaf sync` continues to work exactly as before, with all its existing
flags. `calkit sync overleaf` and `calkit overleaf sync` are the same command
registered on both apps, so there's no duplicated logic.

## Extensibility

Targets register through a small registry (target name → sync callable +
is-configured check), so `sync all` iterates over registered targets rather than
hard-coding them. Future targets like `gdrive` and `zotero` mentioned in the issue
can be added by registering one entry plus a thin subcommand.

## Testing

Added `calkit/tests/cli/test_sync.py` covering the new group's help output, that
`sync git`/`sync dvc` invoke pull-then-push, that `sync overleaf` is accessible, and
that `sync all` runs configured targets while skipping unconfigured ones. Existing
Overleaf tests still pass.

### DIFF
--- a/calkit/cli/main/core.py
+++ b/calkit/cli/main/core.py
@@ -59,6 +59,7 @@ from calkit.cli.notebooks import notebooks_app
 from calkit.cli.office import office_app
 from calkit.cli.overleaf import overleaf_app
 from calkit.cli.scheduler import scheduler_app
+from calkit.cli.sync import register_sync_target, sync_app
 from calkit.cli.update import update_app
 
 app = typer.Typer(
@@ -89,6 +90,7 @@ app.add_typer(
     help="Work with a job scheduler (SLURM or PBS).",
 )
 app.add_typer(dev_app, name="dev", help="Developer tools.", hidden=True)
+app.add_typer(sync_app, name="sync", help="Sync with disparate systems.")
 
 
 def _to_shell_cmd(cmd: list[str]) -> str:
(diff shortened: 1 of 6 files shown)


ChetanyaRathi:
Thank you for the review @petebachant! I've addressed the Copilot comments and pushed fixes, and merged latest main. Ready for another look whenever you get a chance to approve the CI run.

ChetanyaRathi:
Thank you for jumping in and pushing these through, @petebachant — appreciate it. LGTM on my end.

PR #1010, opened 2026-07-05
Capture stage process logs into .calkit/local (closes #1008)

### LINKED ISSUE #1008: Logs from stage processes should be captured
Currently we capture the DVC output, but not the output of the stages themselves, which would probably be good to have, even if they only go into `.calkit/local` when `--log` is not used with the `run` command. We should probably put `systems` and `runs` into `.calkit/local` by default as well, for the sake of local traceability.

ChetanyaRathi:
## What this does

Captures the output of pipeline stage processes, not just DVC's wrapper output (closes #1008).

- **Without `--log`:** each stage's stdout/stderr is captured to `.calkit/local/logs` (gitignored) — one main run log plus one per-stage log. The tracked `.calkit/logs` directory stays empty. Run info and system info are also written under `.calkit/local` (`runs/` and `systems/`), for local traceability without adding tracked state.
- **With `--log`:** logs are additionally copied to the tracked `.calkit/logs` directory, along with run info. `--log` is a boolean flag.

## Also included

- Fixes a regression where `calkit run` raised `ValueError: Reference at 'refs/heads/master' does not exist` in a repo with no commits yet (e.g. right after `calkit init`). The `git_rev = repo.head.commit.hexsha` lookup is now guarded, mirroring the existing detached-HEAD guard on the branch lookup.

## Tests

Adds tests in `calkit/tests/cli/main/test_core.py` covering private-log capture without `--log`, the `--log` copy-to-tracked behavior, and that `.calkit/local` is gitignored with `runs/`/`systems/` created.

## Notes

Verified on Windows locally. A few tests fail only on Windows for environmental reasons unrelated to this change — Device Guard/Application Control blocking `dvc.exe`/`calkit.exe` (WinError 4551), `cat`/`touch`/`true` not existing in `cmd`, DVC lock timing, and `--force` path handling — and pass on Linux CI.

Closes #1008


ChetanyaRathi:
Thank you for the review @petebachant! I've pushed fixes for the Copilot comments — the tracked .calkit/{logs,runs,systems} dirs are now created up front (fixes the FileNotFoundError under --log), the SIGINT restore checks is not None so SIG_DFL isn't dropped, and stage names are sanitized before going into log filenames. I also merged latest main to clear the conflict. The concurrency and per-stage-log retention points are valid but more design-level — happy to take those in a follow-up so this PR stays focused on capture, unless you'd rather they land here. Ready for another look whenever you get a chance to approve the CI run.

calkit/tests/cli/main/test_core.py
```diff
@@ -2029,3 +2034,114 @@ def test_call_dvc_passthrough_hint(tmp_dir):
...
+        "stages": {
+            "test_stage_log": {
+                "cmd": "python stage_script.py",
+            }
+        }
+    }
+    with open("dvc.yaml", "w") as f:
+        import yaml
+
+        yaml.dump(dvc_yaml, f)
```
petebachant: Probably put import at top and perhaps just use `calkit.ryaml`

calkit/tests/cli/main/test_core.py
```diff
@@ -2029,3 +2034,114 @@ def test_call_dvc_passthrough_hint(tmp_dir):
...
+
+    res = subprocess.run(["calkit", "run", "--log"])
+    assert res.returncode == 0
+
```
petebachant: Blank lines to remove in this function

petebachant:
Tested out locally, and `.calkit/local/runs` do not auto-truncate at the last 10 records. They should, however. I think this should probably part of a user's Calkit config, with a key like `run_history_length` or something.

ChetanyaRathi:
Added run_history_length (default 10) to config and generalized the prune helper to cap .calkit/runs too, with a test. One thing to confirm: run info writes to the tracked .calkit/runs, so pruning removes committed JSON on each run past the cap (shows as deletions in git status, picked up by the save step). Applied the cap there per your note — happy to scope it to a gitignored location instead if you'd rather run history not be tracked. The yaml-import and blank-line comments were already resolved in the earlier merge.

petebachant:
> Added run_history_length (default 10) to config and generalized the prune helper to cap .calkit/runs too, with a test. One thing to confirm: run info writes to the tracked .calkit/runs, so pruning removes committed JSON on each run past the cap (shows as deletions in git status, picked up by the save step). Applied the cap there per your note — happy to scope it to a gitignored location instead if you'd rather run history not be tracked. The yaml-import and blank-line comments were already resolved in the earlier merge.

I don't think we want pruning to happen when `--log` is specified, i.e., when the runs go into `.calkit/runs` instead of `.calkit/local/runs`. I believe @rebeccamccabe is logging the run metadata in CI, so pruning there would not be helpful I assume.

ChetanyaRathi:
Good call, i just moved the prune to .calkit/local/runs (gitignored) and left .calkit/runs untouched.

petebachant:
I made a few changes here. Instead of multiple logs per run--one main and one per stage, they are now all interleaved for simplicity, with separator lines to mark the inner versus outer processes. Any thoughts on this UX? I think it still solves with problem but in a simpler way.

PR #1022, opened 2026-07-17
Disable OS credentials store for Overleaf Git ops

### LINKED ISSUE #1021: Unable to reset overleaf_token after it expires
My overleaf git token expired, so I generated a new one. I tried running `calkit config set overleaf_token mytokenhere` but it didn't work. But `calkit config get overleaf_token` correctly gets the new token. I also tried unsetting and then setting, same behavior. Could the old token be cached somewhere even if `calkit config get` is not?
The error I get is `Error: Failed to pull from Overleaf; check that your Overleaf token is valid
Run 'calkit config get overleaf_token' and ensure that it matches one in your Overleaf account settings (https://overleaf.com/user/settings)`

petebachant:
Resolves #1021

### DIFF
--- a/calkit/overleaf.py
+++ b/calkit/overleaf.py
@@ -30,6 +30,59 @@ def get_git_remote_url(project_id: str, token: str) -> str:
     return f"https://git:{token}@git.overleaf.com/{project_id}"
 
 
+# Git config options that force authentication to use only the token embedded
+# in the Overleaf remote URL, never a credential saved in the operating
+# system's credential store (macOS Keychain, Windows Credential Manager, etc.).
+# An expired token can linger in that store and shadow a freshly set one, which
+# otherwise makes it impossible to recover a project after a token expires.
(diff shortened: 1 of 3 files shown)


calkit/overleaf.py
```diff
@@ -30,6 +30,59 @@ def get_git_remote_url(project_id: str, token: str) -> str:
...
+        remote_url,
+        dest,
+        multi_options=_CREDENTIAL_CLONE_OPTIONS,
+        allow_unsafe_options=True,
+    )
+    _disable_credential_store(repo)
+    return repo
+
+
+def clone_or_open(remote_url: str, dest: str | PathLike) -> git.Repo:
```
petebachant:
```suggestion
def get_repo(remote_url: str, dest: str | PathLike) -> git.Repo:
```

petebachant: We could also perhaps omit `dest`, since Calkit projects clone to a fixed, predictable location under the project working directory.

PR #1034, opened 2026-07-29
Speed up a few slow tests

petebachant:
These are some big ones slowing down CI and are probably overly complex.

### DIFF
--- a/calkit/tests/dvc/test_core.py
+++ b/calkit/tests/dvc/test_core.py
@@ -1,17 +1,20 @@
 """Tests for the ``dvc`` module."""
 
+import logging
 import os
 import subprocess
 
 import dvc.repo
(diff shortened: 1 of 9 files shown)


calkit/tests/dvc/test_core.py
```diff
@@ -109,6 +110,46 @@ def test_stale_rwlock_warning_is_suppressed(caplog):
     assert any("must pass through" in m for m in messages)
 
 
+def test_tolerate_lock_release_failures(tmp_dir, caplog):
+    import logging
+
+    import zc.lockfile
+
+    from calkit.dvc.core import _tolerate_lock_release_failures
```
petebachant: These imports can probably go at the top of the module

PR #1040, opened 2026-07-29
Allow stage-scoped system dependencies (#1029)

### LINKED ISSUE #1029: I want to declare a system dependency only for some stages, or declare a system dependency for an environment
I currently have 
```
dependencies:
  - docker
  - matlab
```
but when trying to run a stage that does not require docker or matlab, I still get the error `app 'matlab' not found`. (In my case this is for the AI agent, since I haven't yet figured out how to give it access to a matlab license, but could equally be true of a user who doesn't have a matlab license or doesn't want to bother changing any matlab stages and is just trying to reproduce downstream post processing).

For now I'm just commenting the matlab dependency out, but it would be useful to be able to declare stage-specific or environment-specific system deps so that only the relevant stages are blocked by not having that dependency.

I'm using the matlab system dependency instead of the matlab type environment due to licensing issues described here: https://github.com/calkit/calkit/issues/509.

ChetanyaRathi:
## What this does

Closes #1029.

Right now system dependencies listed under `dependencies:` (like docker or matlab) are checked for the whole project. So a stage that needs neither still fails with an error like `app 'matlab' not found`.

This change lets you scope a system dependency to specific stages, so only the stages that actually need it are blocked when it's missing. Stages that don't need matlab can run even without a matlab license installed.

## Changes
- `calkit/core.py` — dependency-checking logic now respects stage scope
- `calkit/models/pipeline.py` — model updated to allow declaring a dependency per stage
- `calkit/cli/main/core.py` — small wiring change
- `calkit/tests/test_dependencies.py` — tests covering scoped dependencies

## Testing
- `test_dependencies.py` passes locally (10 tests)
- pre-commit passes (ruff, ruff-format, prettier, etc.)
- Didn't run the full suite locally (needs Julia/Conda/Docker/TeX); relying on CI for that

Happy to adjust the approach or the config shape if you'd prefer something different.

### DIFF
--- a/calkit/cli/main/core.py
+++ b/calkit/cli/main/core.py
@@ -1935,7 +1935,9 @@ def run(
     if not quiet:
         calkit.echo("🔗 Checking system-level dependencies")
     try:
-        calkit.check_system_deps(ck_info=ck_info, system_info=system_info)
+        calkit.check_system_deps(
+            ck_info=ck_info, system_info=system_info, targets=targets
+        )
     except Exception as e:
         os.environ.pop("CALKIT_PIPELINE_RUNNING", None)
         raise_error(str(e))
--- a/calkit/core.py
+++ b/calkit/core.py
@@ -445,6 +445,7 @@ def check_system_deps(
     system_info: dict | None = None,
     interactive: bool | None = None,
     use_cache: bool = True,
+    targets: list[str] | None = None,
 ) -> None:
     """Check that the dependencies declared in a project's ``calkit.yaml`` file
     exist.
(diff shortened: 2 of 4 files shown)


petebachant:
I'm not so sure about adding stage level dependencies. I was maybe thinking a `system` environment kind, where dependencies could live there, and then the environment could be referenced by name instead of `_system`. We'd also need to design some sort of system environment lock file as well.

ChetanyaRathi:
That makes sense — a named system environment is cleaner than tagging individual stages, and it fits the existing environment model better than my per-stage approach.

Before I rework this: is the system-environment-kind + lock file something you'd want me to take on, or is it enough of a design decision that you'd rather spec it (or handle it) yourselves? Happy to help, just want to make sure I'm not building ahead of a design you haven't settled. If you do want me on it, could you sketch what the lock file should capture — just the resolved versions of the system deps, or more?

petebachant:
System environments are now merged into `main`, with dependencies backward-compatibly renamed to requirements, and these can be defined either at the project or the system environment level, with an optional `lock` list for properties that should invalidate stages in which the environment is used when they change.

Is this PR now obsolete or is there anything we should merge after resolving conflicts?

PR #1041, opened 2026-07-29
Warn about gitignored files making a stage stale (#1036)

### LINKED ISSUE #1036: I want `calkit status` (or another command) to tell me about ignored files making a stage stale
I had a stage that used a folder as its dependency rather than listing out all individual files inside it. When there was an extraneous temporary git-ignored file in that folder (that I was oblivious to) in my development environment but not in the CI, it caused the stage to rerun every time I went back and forth between a local and a CI run even when there were no meaningful changes.

It was particularly tricky to pin down why because status showed no changes. It would be helpful for either status to notify about such files, or for `calkit pull` to prompt and delete them, or some other way that such changes get revealed during a normal workflow.

ChetanyaRathi:
## What this does

Closes #1036.

When a stage depends on a whole folder (rather than listing individual files), an extra git-ignored file sitting in that folder can quietly make the stage look stale. This is confusing because `calkit status` shows no changes, yet the stage keeps re-running when moving between a local machine and CI.

This change detects that situation and warns the user when a directory dependency contains git-ignored files that could be affecting staleness, so the cause is visible during a normal workflow instead of being a silent mystery.

## Changes
- `calkit/pipeline.py` — added the detection and warning for git-ignored files in directory dependencies
- `calkit/tests/test_pipeline.py` — tests covering the new warning
- `testdata/.gitignore` — test fixture

## Testing
- `test_pipeline.py` passes locally (38 tests)
- pre-commit passes (ruff, ruff-format, prettier, etc.)
- Didn't run the full suite locally (needs Julia/Conda/Docker/TeX); relying on CI for that

Happy to adjust the wording of the warning or where it surfaces if you'd prefer something different.


calkit/tests/test_pipeline.py
```diff
@@ -2394,3 +2394,45 @@ def test_ensure_latex_aux_gitignore(tmp_dir):
...
+    subprocess.check_call(["calkit", "init"])
+    os.makedirs("data")
+    with open("data/ignored.txt", "w") as f:
+        f.write("ignored")
+    with open("data/.gitignore", "w") as f:
+        f.write("ignored.txt\n")
+    # Git add the .gitignore
+    subprocess.check_call(["git", "add", "data/.gitignore"])
+    subprocess.check_call(["git", "commit", "-m", "init"])
+
```
petebachant: Breaking some style rules here with the blank line, per `AGENTS.md`

```suggestion
```

testdata/.gitignore
petebachant, line 1: Test data typically goes in the `test` folder, but this seems simple enough to generate on the fly in conjunction with `tmp_dir`.

ChetanyaRathi: Removed the committed fixture — the test already creates its .gitignore inside tmp_dir at runtime, so nothing extra was needed.

ChetanyaRathi:
This looks already resolved in the current version — I don't see a blank line inside the function body there now. Happy to fix if you're still seeing one.

calkit/pipeline.py
```diff
@@ -962,6 +962,46 @@ def _stage_sort_key(display_name: str) -> tuple[int, int]:
...
+                                "--ignored",
+                                "--exclude-standard",
+                                "--others",
+                                dep_path,
+                            ],
+                            capture_output=True,
+                            text=True,
+                        )
+                        ignored_files = res.stdout.strip().splitlines()
+                        if ignored_files:
```
rebeccamccabe: I think this implementation would trigger the warning even when the gitignored file is a dvc-tracked pipeline output, which would be a false positive. The case to warn for is `if ignored_files not in dvc_tracked_pipeline_outputs`.

rebeccamccabe: or maybe actually `dvc_tracked_files` at all, even if they aren't pipeline outputs? Whichever one would not otherwise show up in status as stale.

ChetanyaRathi:
Fixed, I now skip DVC-tracked files, so the warning only fires for git-ignored files DVC isn't already accounting for. Added a test with a dvc-added file to confirm it doesn't false-positive.

ChetanyaRathi:
Confirmed — the positional path already covers "which target," so I'll drop --target and put the bundling behavior behind a new flag. Good news on the wiring: get_stage_for_output already resolves an output path to its stage, so I can map the release path (e.g. publications/paper.pdf) to the producing stage and bundle its upstream chain from there.

For the flag name I'd lean --self-contained (reads as "rebuildable from itself" = pipeline + dvc.lock + deps), but --full-provenance works too if you'd rather stress the provenance angle. Which do you want me to use?

calkit/tests/test_pipeline.py
```diff
@@ -2394,3 +2394,61 @@ def test_ensure_latex_aux_gitignore(tmp_dir):
...
+                    "inputs": ["data"],
+                }
+            }
+        },
+    }
+    with open("requirements.txt", "w") as f:
+        f.write("requests\n")
+    with open("calkit.yaml", "w") as f:
+        calkit.ryaml.dump(ck_info, f)
+
```
petebachant: More of these

```suggestion
```

calkit/pipeline.py
```diff
@@ -962,6 +962,71 @@ def _stage_sort_key(display_name: str) -> tuple[int, int]:
                     for target in targets
                 )
             }
+
+        import logging
+        import subprocess
+
+        log = logging.getLogger(__name__)
```
petebachant: Can be moved to top of module, or does this need to be inside this function?

calkit/pipeline.py
```diff
@@ -962,6 +962,71 @@ def _stage_sort_key(display_name: str) -> tuple[int, int]:
...
+                    dvc_tracked_paths.add(
+                        Path(out.fs_path)
+                        .relative_to(_dvc_repo.root_dir)
+                        .as_posix()
+                    )
+                except (ValueError, AttributeError):
+                    pass
+        except Exception:
+            pass
+
```
petebachant:
```suggestion
```

calkit/pipeline.py
```diff
@@ -962,6 +962,71 @@ def _stage_sort_key(display_name: str) -> tuple[int, int]:
...
+                            log.warning(
+                                f"Stage '{name}' has a directory dependency '{dep_path}' "
+                                f"that contains git-ignored files (e.g., '{stray_ignored[0]}'). "
+                                "This can cause the stage to unexpectedly re-run because the "
+                                "git-ignored files change the directory's DVC hash. "
+                                "Consider removing them or making the dependency more specific."
+                            )
+                    except Exception:
+                        pass
+
```
petebachant:
```suggestion
```

ChetanyaRathi:
Moved logging and subprocess to the top of the module, there is no reason to keep them inline.

calkit/tests/test_pipeline.py
```diff
@@ -2394,3 +2394,60 @@ def test_ensure_latex_aux_gitignore(tmp_dir):
     assert not os.path.exists(os.path.join("doc", ".gitignore"))
     with open(os.path.join("sub", "doc", ".gitignore")) as f:
         assert "*.aux" in f.read()
+
+
+def test_get_status_warns_gitignored_directory_dependency(tmp_dir, caplog):
+    import subprocess
```
petebachant: I think we can move this import to the top of the module

calkit/tests/test_pipeline.py
```diff
@@ -2394,3 +2394,60 @@ def test_ensure_latex_aux_gitignore(tmp_dir):
...
+    subprocess.check_call(["calkit", "init"])
+    os.makedirs("data")
+    with open("data/ignored.txt", "w") as f:
+        f.write("ignored")
+    with open("data/.gitignore", "w") as f:
+        f.write("ignored.txt\n")
+    # Git add the .gitignore
+    subprocess.check_call(["git", "add", "data/.gitignore"])
+    subprocess.check_call(["git", "commit", "-m", "init"])
+
```
petebachant:
```suggestion
```

calkit/pipeline.py
```diff
@@ -962,6 +964,66 @@ def _stage_sort_key(display_name: str) -> tuple[int, int]:
                     for target in targets
                 )
             }
+
+        log = logging.getLogger(__name__)
+
```
petebachant:
```suggestion
        log = logging.getLogger(__name__)
```

ChetanyaRathi:
subprocess is already imported at module scope (line 4), so I dropped the redundant inline one. Thanks

petebachant:
Testing this out locally, and it does what it says it does, but I wonder if we can make it a bit more user friendly and only warn when an ignored file is making a stage stale. Or, when a stage is stale, warn in the pipeline status section?

Currently we have a blanket print statement at the top of the status output (should use `cli.warn` so it's yellow)

image

What I envision is for this information to go into the pipeline status itself:

```
---------------------------- Pipeline ----------------------------
Stale stages:
        build-paper2:
          modified inputs:
            paper2/figures (WARNING: has ignored files)
```

Any opinions on the best time/place for it to show up @rebeccamccabe?

rebeccamccabe:
I agree that it would be more useful to have it only appear in the status
for the relevant stage.

As for only appearing when it’s making a stage stale, I could see either
way: I might want it even when it’s not currently making a stage stale
because it could make a stage stale when run on CI later.

Maybe the way to go would be to print a list of all such gitignored files
that have changed (making stage draw) and another list of all such
gitignored files that have not changed. Seeing all of them is more useful
than the first one e.g. in the current implementation.

-Becca

ChetanyaRathi:
Sounds good,  folding it into the pipeline status per stage rather than the top-level line. Two things before I wire it up:

The current warning is a log.warning from inside get_status, which is why it shows up as that plain (non-yellow) line above the status block. To put it inside the status per stage it needs to become structured data — a field on StaleStage, plus something at the PipelineStatus level for deps whose stage isn't currently stale — and a change in the status renderer. Slightly more than moving the log line, so want to confirm you're good with that shape.

For the changed-vs-not split: a dep's ignored files are "causing staleness now" when that dep is in the stage's modified inputs, and "not now, but could on CI" otherwise — those are baked into dvc.lock locally and would be absent on a fresh CI checkout, which is the case @rebeccamccabe flagged.

Shape I would propose:

---------------------------- Pipeline ----------------------------
Stale stages:
        build-paper2:
          modified inputs:
            paper2/figures (ignored, changed: notes.tmp)

Ignored files in stage deps (not stale now, may bite on CI):
        build-paper2 / paper2/figures:
            cache.bin, scratch.log

The open bit is where that second block goes — happy to inline it under each stage instead. Let me know if this works and I'll implement.

rebeccamccabe:
| a dep's ignored files are "causing staleness now" when that dep is in the
stage's modified inputs
I think the above statement isn’t true because the modified inputs will
only show the folder, which is exactly why this case is difficult to detect
currently. A dvc md5 entry is only made for the entire folder, not each
file within it. So I believe the correct way to detect if it’s causing
staleness would be process of elimination: if its upper level dvc folder is stale and the ignored file exists and the staleness of the folder
is not being caused by another git tracked or dvc tracked file within the
folder.

-Becca

ChetanyaRathi:
You're right, thanks — the folder-level .dir hash is the whole reason this is hard. modified_inputs containing the dep only tells us the folder's hash moved, not which file inside moved it, so I was wrong to treat that as "the ignored file changed."

So attribution has to be elimination like you said: dep folder stale, stray ignored file present, and the folder's staleness not explained by a git-tracked or dvc-tracked file inside it (dvc-tracked ones I already filter; git-tracked I can detect with git status scoped to the folder). If nothing tracked in the folder moved, the ignored files are the cause.

One thing to pin down first, since it changes the machinery: elimination gives a folder-level answer — "these ignored files are collectively why this folder is stale" — not a per-file changed/not-changed split. The only baseline for whether a specific ignored file changed is the cached .dir manifest behind the dir hash (the files are gitignored and have no separate DVC record), so per-file precision means reading that out of .dvc/cache and diffing md5s — doable but couples us to the cache layout, which moved in DVC 3.x.

Two options:

Folder-level: under each stale stage, list the dep's ignored files when elimination says they're the cause; separately list ignored files in non-stale dep folders as a reproducibility hazard (their state isn't captured in git, so another machine/CI can disagree on the hash). No cache reading.
Per-file: same, plus read the .dir manifest to split each folder's ignored files into actually-changed vs unchanged.

Do you need per-file precision, or is folder-level attribution enough? Happy to build either — just want to do it once.

rebeccamccabe:
My thought would be that folder level is sufficient, although I am not
familiar with the .dir manifest and I suppose file level could be nice if
it’s not any more complex to implement than folder level.

-Becca

ChetanyaRathi:
Great — folder-level it is. Per-file would mean reading the cached .dir manifest out of .dvc/cache and diffing md5s, which ties us to the cache layout (that changed in DVC 3.x), so it's meaningfully more complex than the folder-level check — going folder-level as you suggested. Implementing now.

ChetanyaRathi:
Implemented, folder-level as agreed. Summary:

The warning is now structured data instead of a top-level log line: StaleStage.ignored_stale_deps for the attributed-cause case (shown inline under a stale stage's modified inputs), and PipelineStatus.ignored_dep_hazards for ignored files in dep folders that aren't currently causing staleness but could on CI. Rendered via cli.warn (yellow) in the status output.
Attribution is by elimination per your suggestion, @rebeccamccabe: a dir dep is stale, holds stray ignored files, and git status on the folder shows no other change → the ignored files are named as the cause. If a tracked file in the folder also changed, we defer to that and demote the ignored files to the hazard list rather than claim them as the cause. Tests cover that false-positive guard explicitly.
Dropped the now-unused logging import since the warning is no longer logged.

petebachant:
Tested against the latest pushed here--looks pretty cool!

image

in the JSON, we should rename `ignored_stale_deps` to `ignored_modified_inputs` to be consistent, however:

```json
    "stale_stages": {
      "build-paper2": {
        "raw_status": [
          {
            "changed deps": {
              "paper2/figures": "modified"
            }
          }
        ],
        "stale_outputs": [],
        "modified_inputs": [
          "paper2/figures"
        ],
        "modified_outputs": [],
        "modified_command": false,
        "always_run": false,
        "ignored_stale_deps": {
          "paper2/figures": [
            "paper2/figures/ignored.txt"
          ]
        },
        "is_subproject": false
      }
    },
```

petebachant:
Took a look at how this will show up when the pipeline is up-to-date but there are ignored deps in there:

image

What do we want the user to do in response to this? Should they add the files to `.dvcignore`, remove the ignored files and rerun. It is potentially valid to have an upstream stage create an ignored file that a downstream stage uses, but I can see how CI will make the downstream run again if the upstream doesn't, though if it's declared as an output, the upstream should see it changed and rerun.

rebeccamccabe:
It would probably be useful eventually for there to be an option to make
the pipeline automatically delete such files to prevent the staleness, for
the case where the output is not needed for the downstream stage but is ie
a log created by a simulation and not worth tracking, but doing this
automatically seems like a separate feature. The current feature would now
alert to those files existence and in response the user would manually run
a git clean if they don’t want those files around.

I currently don’t list such log files as outputs of the stage because I
don’t want the first stage to rerun when the log is missing or changed (ie
a timestamp).

-Becca

ChetanyaRathi:
Renamed ignored_stale_deps → ignored_modified_inputs and pushed — agreed, sitting next to modified_inputs it reads as the ignored subset of that. Left ignored_dep_hazards as-is since those are files in non-stale deps, so the modified-inputs parallel doesn't apply — happy to rename it too if you'd prefer a matching pattern.

On what the user should do: agree with @rebeccamccabe that manual git clean (or .dvcignore if they want to keep the files) is the right response, and auto-deletion is a separate feature rather than something this PR should do. The one gap is the hazard block reports the situation without pointing at the fix — want me to add a short "run git clean or add to .dvcignore" hint after the hazard list, or keep the output terse and put that in the docs? Either's quick.

ChetanyaRathi:
Since this keeps the output terse, I'll default to a one-line pointer under the hazard block, something like → run 'git clean' to remove, or declare them dvc-ignored (#1053) once that lands. Happy to drop it or move it to docs instead. If that is fine then i will add it in a follow-up commit; otherwise this is ready as it is.

rebeccamccabe:
Is this ready to merge? (asking because I just had another hidden gitignored file that was making my stage stale and this feature would be useful!)

PR #1051, opened 2026-08-04
Add calkit.yaml schema, system environment kind, and rationalize SSH with envs

### LINKED ISSUE #1043: I want the docs site to include available properties for inputs/outputs
Available values for pipeline parameters that get nested under `inputs` and `outputs`, ie `path`, `storage`, are currently not part of the API reference on the pipeline page.

### LINKED ISSUE #1531: Drop `SSHEnvironment`
The model doesn't really make sense. It should be a `system` environment that has a host, key, etc., and if we're already there, we don't need to SSH into it. Otherwise, we need a workspace on the system to run in. Maybe an operator connected to a hub is a better model.

petebachant:
## TODO

- [x] Remove `_include` support.
- [x] Confirm this resolves #1043.
- [x] Make this resolve #1531.
- [x] System properties to lock--should they be under the `lock` key, or `props` or something else, indicating the properties of the system we care about? Maybe unify `dependencies` and `lock` to be `requirements`.


calkit/models/core.py
```diff
@@ -95,127 +131,225 @@ class Environment(BaseModel):
...
 class JuliaEnvironment(Environment):
     kind: Literal["julia"] = "julia"
-    path: str
-    julia: str
+    path: str = Field(description="Path to the Julia project's Project.toml.")
+    julia: str = Field(description="Julia version to use.")
 
 
 class MatlabEnvironment(Environment):
     kind: Literal["matlab"] = "matlab"
```
petebachant, line 537: Maybe worth renaming to `matlab-docker` to be specific. Maybe we'll have `mip` environments in here someday, though those probably don't include toolboxes.

calkit/models/core.py
```diff
@@ -50,14 +58,19 @@ class Presentation(_CalkitObject):
...
-        "blog",
-    ]
+    # Optional since publications can be created without one, e.g., by
+    # ``calkit overleaf sync``, whose ``--kind`` option has no default
+    kind: (
+        Literal[
+            "journal-article",
+            "conference-paper",
+            "proposal",
+            "poster",
```
petebachant: Some irrationality here. Is a poster a publication? Maybe we need a posters section, like presentations?

calkit/models/core.py
petebachant: Or maybe we use `system-matlab` for MATLAB installed at the OS level.

calkit/environments.py
```diff
@@ -399,6 +430,15 @@ def get_env_lock_fpath(
...
         lock_fpath = os.path.join(env_dir, "flake.lock")
+    elif env_kind == "system":
+        # A system env's "lock" records the machine properties it declared it
+        # depends on. With nothing locked there is nothing to depend on, so
+        # no lock file and no DVC dep.
+        if not env.get("lock"):
+            return None
+        lock_fpath = os.path.join(env_lock_dir, env_name, "info.json")
+        if for_dvc:
+            lock_fpath = os.path.dirname(lock_fpath)
```
petebachant: The full path for the lock file is probably the right choice here, not the directory.

calkit/cli/main/core.py
```diff
@@ -2634,6 +2649,28 @@ def run_in_env(
...
+        list[str] | None,
+        typer.Option(
+            "--get",
+            help=(
+                "Path the command writes, to be collected from the "
+                "workspace afterwards. Only used when the environment runs "
+                "on another machine. Normally set by the pipeline."
+            ),
+        ),
+    ] = None,
```
petebachant: Do we still use these?

PR #1068, opened 2026-08-07
Add `max_concurrent_jobs` to scheduler environments

petebachant:
## Claude's decription

An iterate_over stage on a SLURM or PBS environment submits every one of its jobs at once. That is fine when you have the cluster to yourself, but on a shared cluster a long sweep fills the queue and pushes everyone else's jobs behind it.

Setting max_concurrent_jobs on the environment caps how many of the project's jobs may be queued or running at a time; submissions past the cap wait for a slot. Only jobs Calkit submitted for the current project count, so the cap bounds one pipeline's share of the queue rather than everything the user has running.

Two details worth noting:

- The check-for-a-slot-then-submit sequence is serialized with a SQLite IMMEDIATE transaction, since the parallel batch processes that fan out an iterated stage would otherwise all see the same free slot and submit together, which is the exact thing the cap exists to prevent. The lock is held across the submit and the record write, so a concurrent waiter counts the new job before deciding it has room.

- Queue occupancy is answered with one `squeue --me` call rather than one per job, and the key is excluded from the environment lock file: the cap changes when a job is submitted, never what it computes, so turning it on partway through a long sweep must not invalidate the cases that already finished.


calkit/tests/cli/test_scheduler.py
```diff
@@ -455,3 +457,134 @@ def test_record_job_result(tmp_dir):
...
+    # If squeue fails we fall back to polling each job rather than reporting
+    # an empty queue we never confirmed---which would let submissions through.
+    calls.clear()
+    result.update(returncode=1, stdout="", stderr="squeue: error: timeout\n")
+    assert _active_job_ids("slurm", ["1", "2"]) == {"1", "2"}
+    assert len(calls) > 1
+
+
+def test_count_queued_jobs(tmp_dir, monkeypatch):
+    import calkit.cli.scheduler as sched
```
petebachant: Just add this import at the top of the module

docs/hpc.md
```diff
@@ -155,6 +156,33 @@ succeeded are skipped and only the failed one is resubmitted.
 the cache; just editing a script invalidates the affected cases without
 `--force`, so they still re-run concurrently.)
 
+## Limiting queue usage
+
+Submitting every iteration at once is fine when you have the cluster to
+yourself, but on a shared cluster a long sweep can fill the queue and push
```
petebachant: This is only true if there isn't some fairness algorithm in the scheduler.

PR #1501, opened 2026-08-10
Better handle git tags for auto-versioning and fix stale front end chunks

petebachant:


.github/workflows/deploy-hub-production.yml
```diff
@@ -106,3 +106,26 @@ jobs:
           fetch-depth: 0
       - run: docker compose -f docker-compose.yml --project-name ${{ vars.STACK_NAME }} build
       - run: docker compose -f docker-compose.yml --project-name ${{ vars.STACK_NAME }} up -d --remove-orphans
+      - name: Verify the deployed version
```
petebachant, line 109: Not so sure we want this here

PR #1535, opened 2026-08-14
Support multiple apps and add `marimo-html-wasm` stage kind

### LINKED ISSUE #57: I want to be able to use marimo notebooks as part of my project
Definitely needs exploration, but this looks like a cool tool: https://github.com/marimo-team/marimo

They are currently not able to be rendered on GitHub, which makes sharing difficult: https://github.com/marimo-team/marimo/discussions/2868

We could add a feature to render on the project notebooks page.

One value-add we could provide is to make cells reactive to external file changes, and ensure the outputs are cached with DVC.

Another is an enforcement that a virtual environment is created, used, and saved to the project fold for others to use. marimo actually will prompt to install packages that are not installed, but we'd want to ensure these changes are persisted/committed.

Relevant for UX: https://www.reddit.com/r/marimo_notebook/comments/1rwf577/suggestion_make_marimo_notebooks_executable/

### LINKED ISSUE #354: I want to be able to ship multiple apps with my project so users can make predictions for different situations
Some projects may have may predictions to make, so they could have multiple apps. It wouldn't be too hard to allow this. Some will be web apps, and some might not be, and we can host the former.

petebachant:
Resolves #57, resolves #354

## TODO

- [ ] Apps default URL on hub redirects to first app in map?
- [x] Get CLI command and stage schema correct. Is `calkit nb export-app` the right command, and does it fit with marimo, JupyterLite, Panel, etc., or can those simply use `calkit xenv`? Marimo needs our new `--include` command to copy data files into the build dir. `calkit app from-marimo`?
- [ ] Apps tree sidebar should be collapsible?


calkit/models/core.py
```diff
@@ -321,6 +322,36 @@ class Release(BaseModel):
...
+    # record their provenance.
+    stage: str | None = None
+    # Catch typos, and reject a hand-written ``url`` rather than silently
+    # ignoring it
+    model_config = ConfigDict(extra="forbid")
+
+    @property
+    def serve_dir(self) -> str:
+        """The directory to serve, i.e., the app's root."""
+        return PurePosixPath(self.path).parent.as_posix()
```
petebachant, line 377: `self.path` must not be `.`

PR #1557, opened 2026-08-19
fix(hub): highlight Apps in sidebar on app sub-routes

### LINKED ISSUE #1550: Apps is not highlighted in sidebar when on project apps page

sloemo01:
Fixes #1550

## Problem
The **Apps** item in the project sidebar is not highlighted when viewing a specific app page (e.g. `/apps/my-app`).

## Root cause
`SidebarItems.tsx` uses `activeOptions={{ exact: true }}` for every nav item. The Apps section is the only one with sub-routes (`/apps/<appName>`), so on a specific app page the path is not exactly `/apps` and the item loses its highlight.

## Fix
Set `exact: false` for the Apps item only, so it stays highlighted on any `/apps/*` path while all other items keep exact matching.

Verified with `tsc --noEmit` (passes).

### DIFF
--- a/hub/frontend/src/components/Common/SidebarItems.tsx
+++ b/hub/frontend/src/components/Common/SidebarItems.tsx
@@ -107,6 +107,10 @@ const SidebarItems = ({ onClose, basePath }: SidebarItemsProps) => {
       return null
     }
 
+    // The Apps section has sub-routes (/apps/<appName>), so it should stay
+    // highlighted on any of them, not just the exact /apps path.
+    const isApps = path === "/apps"
+
     return (
       <Flex
         key={title}
@@ -116,7 +120,7 @@ const SidebarItems = ({ onClose, basePath }: SidebarItemsProps) => {
         search={currentRef ? ({ ref: currentRef } as any) : undefined}
         w="100%"
         p={2}
-        activeOptions={{ exact: true, includeSearch: false }}
+        activeOptions={{ exact: !isApps, includeSearch: false }}
         activeProps={{
           style: {
             background: bgActive,


sloemo01:
Rebased onto `main` (`d466003f`). The branch had gone stale against the 2.0 hub work and carried one conflict in `SidebarItems.tsx`.

## The conflict

Main had since replaced the `const item = ...` / `TipBubble` structure this branch was written against, while keeping the line the PR actually changes. My contribution is only the `isApps` local plus `exact: !isApps`:

```tsx
// The Apps section has sub-routes (/apps/<appName>), so it should stay
// highlighted on any of them, not just the exact /apps path.
const isApps = path === "/apps"
```

So the resolution keeps that and main's simpler `return (`. The diff is still the same 5 added / 1 removed, nothing else carried in.

## Verified

```
tsc --noEmit -p tsconfig.json     exit 0
biome check SidebarItems.tsx      clean, no fixes
```

I ran the typecheck twice, because my first attempt ran against a dangling `node_modules` symlink and passed for the wrong reason. Worth stating what the second run actually proves rather than just the exit code: I mutated the changed line to `const isApps: number = path === "/apps"`, and `tsc` failed at `SidebarItems.tsx(116,11)` with `TS2322`. The compiler is reading this file, so the clean run is a real result and not an empty include list. Mutation reverted, tree byte-identical to the pushed commit.

CI on the new head: 5 success, 5 skipped, no failures.

One thing I want to flag rather than let you find: this repo's `hub/AGENTS.md` says commits and pushes are for humans, not agents. I read that file after I had already pushed the rebase. The branch is rebased and I am not touching it further without a word from you.

PR #1558, opened 2026-08-19
fix(cli): skip DVC push in 'commit --push' when no DVC files changed

### LINKED ISSUE #380: `calkit commit --push` should not try to push to DVC if no DVC files were added
This is just a convenience thing. I like to call `git commit -am ... && git push`, so `calkit commit -pam ...` is a nice shortcut, but if I only committed to Git, I don't need to push to DVC, especially if the project doesn't use DVC.

sloemo01:
Fixes #380

## Problem
`calkit commit --push` always runs a DVC push after committing, even when the commit only touched Git-tracked files. This wastes time pushing nothing, and is especially annoying for projects that don't use DVC.

## Fix
After committing, inspect the files in the new HEAD commit. If none of them are DVC-tracked (a `.dvc` pointer, `dvc.yaml`, `dvc.lock`, or a dvc-zip pointer under `.calkit/zip/`), skip the DVC push and only push to Git.

## Verification
Manually tested:
- Committing a Git-only change → DVC push skipped (no 'Pushing to DVC remote').
- Committing a DVC-tracked file (large file → `bigdata.bin.dvc`) → DVC push attempted.

### DIFF
--- a/calkit/cli/main/core.py
+++ b/calkit/cli/main/core.py
@@ -1073,7 +1073,21 @@ def commit(
         cmd += ["-m", message]
     subprocess.call(cmd)
     if push_commit:
-        push()
+        # Only push to DVC if the commit actually touched DVC-tracked files
+        # (a .dvc pointer, dvc.yaml, dvc.lock, or a dvc-zip pointer). If the
+        # commit only changed Git-tracked files, skip the DVC push so we
+        # don't waste time pushing nothing.
+        repo = calkit.git.get_repo()
+        committed_files = repo.git.show(
+            "--name-only", "--format=", "HEAD"
+        ).split()
+        touched_dvc = any(
+            f.endswith(".dvc")
+            or f in ("dvc.yaml", "dvc.lock")
+            or f.startswith(".calkit/zip/")
+            for f in committed_files
+        )
+        push(no_dvc=not touched_dvc)
 
 
 @app.command(name="save|sv")


sloemo01:
Rebased onto current `main` (`edf7a235`), which clears the `BEHIND` state that was holding this one — the branch was 174 commits back.

The rebase applied as one commit with no conflict, and the change is byte-identical afterwards. One thing I fixed while rebasing: my own patch had left an extra blank line before `@app.command(name="save|sv")`, so the file is now `ruff` clean on both `check` and `format --check`. That blank line was not from the original PR, it was noise the rebase surfaced.

## Why it was worth re-running rather than just force-pushing

I measured the DVC decision on both sides of it, in a throwaway repo with a bare remote:

```text
git-only commit --push
  [main 2fcd1f1] git-only commit 2
  Pushing to Git remote
  To /tmp/calkit-remote.git
     24ba4ff..2fcd1f1  main -> main          <- no DVC push

commit --push where the commit touched data.csv.dvc
  Warning: No DVC remotes configured; skipping DVC push
  Pushing to Git remote                      <- the DVC path was reached
```

That second line is the control: the DVC branch still runs when the commit really did touch a pointer, and only the wasted push is gone.

`calkit commit`'s own test (`calkit/tests/cli/main/test_core.py::test_commit`) passes.

## On the wider suite, so nobody reads a green claim into this

`calkit/tests/cli/main/` reports **72 failed, 29 passed** on my branch — and the same **72 failed, 29 passed** on a clean `upstream/main` with my change absent. They are `FileNotFoundError` on `docker`, `julia` and `R` in this environment, not regressions. I am reporting the number rather than the summary because the two runs matching exactly is the evidence, and CI is where that suite actually runs.

PR #1559, opened 2026-08-19
Don't silently pass unparseable versions when checking Conda envs

### LINKED ISSUE #205: `conda._check_single` fails with a version like `9e`
I encounters a `jpeg` conda package with this version that caused this issue.

sloemo01:
Fixes #205

## Problem
`_check_single` treated any actual version that failed to parse (e.g. conda's `9e`) as satisfying the requirement, returning `True` even when the installed version did not match a pinned constraint. A wrong version slipped through the environment check and only failed later when the stage actually ran.

## Repro
`_check_single('jpeg=1', 'jpeg=9e', conda=True)` returned `True` (should be `False`).

## Fix
An unparseable actual version is now only accepted when the requirement has **no** version constraint (a bare package name). If the requirement pins a version (`=`, `==`, `>`, `<`, `!`, `~`), the check returns `False` since we can't confirm a match.

## Tests
Added cases to `test_check_single`:
- `jpeg` vs `jpeg=9e` (no constraint) → `True`
- `jpeg=1` vs `jpeg=9e` → `False`
- `jpeg>=1` vs `jpeg=9e` → `False`


calkit/conda.py
```diff
@@ -279,10 +279,21 @@ def _check_single(
     try:
         version = Version(actual_vers)
     except InvalidVersion:
+        # An unparseable actual version (e.g. conda's "9e") can't be compared
+        # against a version constraint. If the requirement pins a specific
+        # version we can't confirm it matches, so treat it as not satisfied
+        # rather than silently passing. Only a bare package name with no
+        # version constraint is safe to accept, since any version is allowed.
```
petebachant: Couldn't we properly compute `==` and `!=`?

sloemo01: Good call. Pushed c0bfa09.

Equality and inequality don't need version ordering, so those are now decided by comparing the version strings directly instead of bailing out. `jpeg=9e` against an installed `jpeg=9e` returns True now, which is the case that was wrong before. Ordering operators (`<`, `<=`, `>`, `>=`, `~=`) and wildcards still need a parsed version, so those stay unverifiable and keep returning False with a warning.

While I was in there I noticed the package name was split on `[=<>]` only, which misses `!` and `~`. So `jpeg!=1.0` parsed as package `jpeg!` and then failed the name comparison, meaning every `!=` and `~=` requirement came back False no matter what was installed. Both splits include `!` and `~` now, with tests.

One limitation worth flagging: the match is an exact string compare, so `==1.0` won't match an installed `1.0.0`, since there's no parsed version to zero-pad from. Happy to handle that too if you think it matters.

PR #1560, opened 2026-08-19
Add `markdown` stage kind

### LINKED ISSUE #671: I want to be able to define a pipeline and environments in Markdown, so I can have a "runnable README"
We could add to the line where language would typically be defined, e.g.,

````md
```sh calkit stage --environment=main
python my-script.py
```
````

Environments could be declared in a similar way:

````md
```sh calkit environment --name=main --kind=uv-venv --path=requirements.txt
pandas==1.0
requests
```
````

This is basically just like the notebook stage magic. It's also sort of overlapping with Quarto's functionality, but this is a bit simpler. Not really made for creating publications, but making READMEs runnable.

On the other hand, this could be a bit confusing since the README will be telling users to manually run these commands, but they could run them all automatically with `calkit run`. It would be better to point to the script rather than list the command used to run the script, maybe like:

>The script used to generate the figures is `scripts/plot.py`.
>It uses the `main` environment, which is a `uv-venv` using Python version 3.14 with the following packages: `pandas` and `requests`.
>It reads `data/data.csv` and produces a set of figures in the `figures` directory.

However, how do we deterministically parse this prose into environment and pipeline stages?

## Examples

- https://github.com/adam-s/testing-claude-agent

### PR DESCRIPTION
**Pull request by petebachant**
Resolves #671 

## TODO

- [x] Output injection?
- [x] Docs and tutorial
- [x] Check for performance regressions. This adds compilation steps.
- [ ] ~~`target_path` --> `markdown_path`?~~
- [x] Parse package install blocks as environments, e.g., `pip install ...`, `Pkg.add...`.
- [x] Ensure `calkit xr README.md` works properly when there are no fence annotations.
- [x] How to handle dev packages, i.e., to test the ones dev'd inside the repo? Could be part of environment declarations.
- [ ] Add a GitHub Action so users can ensure their README examples are runnable in CI, fails if code examples don't work, perhaps optionally commits changes that inject outputs back in, but probably not by default. This is for projects that don't really want to become Calkit projects, but want to test their READMEs. Make sure there isn't already something that does this.
- [ ] Rewrite copy by hand

### CODE REVIEW

File: calkit/cli/main/xr.py [hunk 1]
```diff
@@ -11,10 +11,151 @@
...
+    import calkit.markdown
+
+    if not os.path.isfile(markdown_path):
+        raise_error(f"{markdown_path} does not exist")
+    with open(markdown_path, encoding="utf-8") as f:
+        original_text = f.read()
+    text = original_text
+    try:
+        blocks = calkit.markdown.parse_markdown(text, path=markdown_path)
+        specs = calkit.markdown.extract_stages(blocks, markdown_path)
```
petebachant: This should be one function that returns a Pydantic model

calkit/cli/main/xr.py
```diff
@@ -11,10 +11,151 @@
...
+        ck_info = calkit.load_calkit_info()
+    declared_envs = calkit.markdown.extract_environments(blocks, markdown_path)
+    existing_env_names = list(ck_info.get("environments", {}) or {})
+    existing_env_names += list(declared_envs)
+    detected_envs, assignments = calkit.markdown.detect_environments(
+        specs,
+        markdown_path,
+        existing_env_names=existing_env_names,
+        default_env=environment,
+    )
```
petebachant: This also could probably be a single function call to extract environments, detected versus declared with annotations

calkit/cli/new.py
```diff
@@ -32,6 +32,37 @@ def _check_path_dir(path: str):
...
+
+def _readme_sources_pipeline(ck_info: dict) -> bool:
+    """Return whether README.md declares any of the project's stages."""
+    stages = ck_info.get("pipeline", {}).get("stages", {})
+    if not isinstance(stages, dict):
+        return False
+    for stage_name, cfg in stages.items():
+        if not isinstance(cfg, dict) or cfg.get("kind") != "markdown":
+            continue
+        if str(cfg.get("target_path") or stage_name) == "README.md":
```
petebachant: Stage name being `README.md` is probably not a good signal

calkit/cli/new.py
```diff
@@ -32,6 +32,37 @@ def _check_path_dir(path: str):
...
+    if "://" in template or "github.com" in template:
+        return git_url, None
+    parts = template.strip("/").split("/")
+    if len(parts) <= 2:
+        return git_url, None
+    subdir = "/".join(parts[2:])
+    return git_url.rsplit("/" + subdir, 1)[0], subdir
+
+
+def _readme_sources_pipeline(ck_info: dict) -> bool:
```
petebachant: Why do we need this?

calkit/cli/new.py
```diff
@@ -405,16 +467,20 @@ def new_project(
         with open(os.path.join(abs_path, "calkit.yaml"), "w") as f:
             ryaml.dump(ck_info, f)
         calkit.schema.ensure_modeline(os.path.join(abs_path, "calkit.yaml"))
-        # Update README
+        # Update README, unless the pipeline is declared inside it, in
+        # which case overwriting it would delete the project
```
petebachant: I think we need a better mechanism to determine how much of a template's README to save versus discard. Maybe we keep it all and simply change the title and add the description.

calkit/cli/new.py
```diff
@@ -405,16 +467,20 @@ def new_project(
...
-            project_title=title,  # type: ignore
-            project_description=description,
-        )
-        with open(readme_fpath, "w") as f:
-            f.write(readme_txt)
+        if _readme_sources_pipeline(ck_info):
+            typer.echo("Keeping README.md, which declares pipeline stages")
+        else:
+            typer.echo("Generating README.md")
+            readme_txt = calkit.make_readme_content(
```
petebachant, line 471: Does the hub's new project post do this too?

calkit/dvc/core.py
```diff
@@ -485,6 +485,66 @@ def get_dvc_lock_holder(wdir: str | None = None) -> dict | None:
     return {"pid": pid, "cmd": cmd}
 
 
+def dvc_init_needs_subdir(wdir: str | None = None) -> bool:
+    """Return whether ``dvc init`` here needs ``--subdir``.
+
+    DVC refuses to initialize inside an existing Git repository unless it
+    is told the project is a subdirectory of one, which is the case for a
+    self-contained example living inside a larger repo.
```
petebachant: Why doesn't DVC automatically assume subdir? So it can have a hierarchy of pipelines?

calkit/dvc/core.py
```diff
@@ -485,6 +485,66 @@ def get_dvc_lock_holder(wdir: str | None = None) -> dict | None:
...
+    root = repo.working_tree_dir
+    if root is None or os.path.realpath(root) == os.path.realpath(base):
+        return False
+    try:
+        return bool(repo.ignored(base))
+    except Exception:
+        return False
+
+
+def dvc_init_args(wdir: str | None = None) -> list[str]:
```
petebachant:
```suggestion
def make_dvc_init_args(wdir: str | None = None) -> list[str]:
```

calkit/models/pipeline.py
```diff
@@ -1752,6 +1753,37 @@ def dvc_cmd(self) -> str:
...
+        default="_system",
+        description="Environment used by blocks that don't name one.",
+    )
+
+    @property
+    def markdown_path(self) -> str:
+        path = self.target_path if self.target_path is not None else self.name
+        if path is None:
+            raise ValueError("Markdown stage has no path")
+        return Path(path).as_posix()
```
petebachant: I think we need path to be set--don't assume stage name is path, i.e. `target_path` can't be `None`.

calkit/models/pipeline.py
```diff
@@ -1796,6 +1829,29 @@ def set_stage_names(self) -> Pipeline:
             stage.name = stage_name
         return self
 
+    @model_validator(mode="after")
```
petebachant: Shouldn't need this if target path is set.

calkit/models/pipeline.py
```diff
@@ -1977,21 +2033,26 @@ def convert_sbatch_stages(self) -> dict[str, dict]:
             converted[name] = calkit_yaml_stage
         return converted
 
-    def ensure_env_lock_paths_are_inputs(
-        self, env_lock_fpaths: dict[str, str]
+    def ensure_env_paths_are_inputs(
```
petebachant: We don't necessarily need to rename this. We can consider `.python-version` part of the env lock, since it locks the Python version.

calkit/tests/cli/main/test_core.py
```diff
@@ -2341,3 +2341,97 @@ def test_calkit_env_no_longer_selects_a_hub(tmp_dir, capsys):
...
+
+def test_run_refuses_outside_a_project(tmp_dir):
+    """``calkit run`` must not initialize anything outside a project.
+
+    It initializes Git and DVC as needed, so a wrong working directory
+    would otherwise scatter a .dvc directory into an unrelated folder ---
+    and inside an existing Git repo that now succeeds, since DVC is told
+    the project is a subdirectory.
+    """
+    import subprocess
```
petebachant: This isn't already imported at the module level?

calkit/cli/main/core.py
```diff
@@ -2163,7 +2349,7 @@ def run(
     except Exception:
         if not quiet:
             typer.echo("Initializing DVC repo")
-        result = calkit.dvc.run_dvc_command(["init"])
+        result = calkit.dvc.run_dvc_command(calkit.dvc.make_dvc_init_args())
```
petebachant: We could have a new function for this:

```suggestion
        result = calkit.dvc.init()
```

docs/pipeline/markdown.md
```diff
@@ -0,0 +1,331 @@
...
+```python calkit stage name=analysis environment=py outputs=[results.json]
+json.dump({"rms": rms, "n": t.size}, open("results.json", "w"))
+```
+````
+
+and the prose refers to values in it by key, between a pair of markers
+that are invisible when rendered:
+
+```md
+The RMS is <!-- calkit value key=rms path=results.json format="{:.4f}" -->0.2933<!-- /calkit value -->
```
petebachant, line 253: How does `value` potentially conflict with `kind: result` question evidence? `value` seems more general and more correct in this case--there's no reason it must be a result, but I'm a little wary of expanding the number of entities in the data model.

Issue #1545, opened 2026-08-17
Docker and `venv` environments should create cross-platform locks if possible


petebachant:
Can we use Docker/pip to solve the specs and lock for as many platforms as possible? This would eliminate the issue where moving to a new platform invalidates a bunch of stages. However, we may be then implying that the results are platform-independent, which may not be the case.

petebachant:
Docker does this now when possible.

PR #1578, opened 2026-08-25
Integrate Docker registries better with locally built images

### LINKED ISSUE #117: I want it to be easy to add a Docker build GitHub Actions workflow to my project so images are built and ready to pull
Our image tag would then need `ghcr.io` in it, and our stage command should be something different.

We could possibly add a push Docker command when checking the env if it needs to be rebuilt.

### LINKED ISSUE #201: Archive Docker images on Zenodo or Figshare to remove dependency on images that might be deleted someday
https://github.com/pauleve/donodo

### LINKED ISSUE #384: I want my Docker images to be cached in the cloud so they don't need to be rebuilt, just in case dependencies change
When we check a Docker environment, we could pull/push to `ghcr.io` by default, adding that prefix to the image tag automatically. We should probably rely on its SHA256 to identify it.

We could also allow for a `registry` property in a Docker env. By default this is empty, which implies Docker Hub.

We could also save them in DVC inside `.calkit/container-images`, so we can push and pull from there. A project release would also include those.

Should image pushing and pulling also happen in `calkit push` and `calkit pull`? Or maybe pulls happen in env checks?

### LINKED ISSUE #754: Docker envs should re-lock after a change to their image

petebachant:
Resolves #754, resolves #201, resolves #117, towards #1545, resolves #384

Now, a `calkit push` will push a project's locally built images to a registry, if defined. They will also be included in releases, so someone downloading a project from, e.g., Zenodo, will have the images if the registry goes away.

The Docker lock file schema changes here, which will invalidate stages run with older versions. There probably aren't too many people out there doing heavy calculations in Docker, though it should be possible to DVC commit them if the image digest truly didn't change.

## TODO

- [x] Calkit GitHub App should request packages write permissions rather than having users generate PATs and save in `github_packages_token`?
- [x] `RepoDigests` needs to be stable across systems so we don't invalidate stages. Double check. Does the registry URL need to make it in there? The same image in a different registry should be fine enough for reproducibility.


calkit/releases.py
```diff
@@ -538,3 +540,153 @@ def check_project_release_archive(
...
+    to fetch an image back out of a release when no registry can serve it.
+    """
+    ck_info = calkit.load_calkit_info(wdir=wdir)
+    resp = {}
+    for env_name, env in (ck_info.get("environments") or {}).items():
+        if not isinstance(env, dict) or env.get("kind") != "docker":
+            continue
+        image = calkit.docker.get_image_name(env, env_name)
+        if not image:
+            continue
```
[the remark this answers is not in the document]
petebachant: Interesting point. For pure reproducibility concerns, archiving something like an Ubuntu image is probably a good idea since image registries can change, but maybe it is bloat.

calkit/cli/new.py
```diff
@@ -1049,8 +1049,36 @@ def new_docker_env(
...
     ] = None,
+    registry: Annotated[
+        str | None,
+        typer.Option(
+            "--registry",
+            help=(
+                "Registry prefix to push built images to and pull them from "
+                "instead of rebuilding, e.g., 'ghcr.io/someone/some-project', "
+                "or 'auto' for the project's GitHub Container Registry "
+                "namespace."
```
petebachant: 'ghcr.io' is also valid to automatically use the project's namespace.

calkit/cli/new.py
```diff
@@ -1049,8 +1049,36 @@ def new_docker_env(
...
+            ),
+        ),
+    ] = None,
+    platforms: Annotated[
```
petebachant:
```suggestion
    build_platforms: Annotated[
```

calkit/cli/new.py
```diff
@@ -1138,6 +1166,10 @@ def new_docker_env(
         env["layers"] = layers  # type: ignore
     if platform:
         env["platform"] = platform
+    if registry:
+        env["registry"] = registry
+    if platforms:
+        env["platforms"] = platforms  # type: ignore
```
petebachant: This is probably less confusing than having `platform` and `platforms`. `platform` is for running, and `build_platforms` is for building.

```suggestion
        env["build_platforms"] = platforms  # type: ignore
```

calkit/cli/update.py
```diff
@@ -872,16 +872,69 @@ def update_docker_env(
         str | None,
         typer.Option("--image", help="Docker image name/tag."),
     ] = None,
+    registry: Annotated[
+        str | None,
+        typer.Option(
+            "--registry",
+            help=(
+                "Registry prefix to push images to and pull them from, or "
+                "'auto' for the project's GitHub Container Registry "
```
petebachant: 'auto' is a little ambiguous. Someday Calkit might have a registry, and we'd want to use that for `auto`, and that could be confusing. Better to be explicit.

```suggestion
                "'ghcr.io' for the project's GitHub Container Registry "
```

Issue #728, opened 2026-02-26
I want to be able to import/link/sync any path between two repos


petebachant:
The use case I have in mind is merging changes into a config file from the main dev repo to a deployment repo, but this could also be used for imported artifacts. Basically we can use the logic we use for Overleaf syncing: compute and apply diffs and track the latest.

This is sort of like forking a single path from a Git repo into another and being able to merge from upstream.

Another use case is where a research project produces an artifact that would be nice to include in the blog. The syncing information can be contained in the research project and it can publish to a specific set of paths in the blog. In this case, it should probably be one-way.

This could be part of a more generic storage concept, which could involve many different storage providers for a project unified into a single "repo".

```yaml
storage:
  git1:
    kind: git
    repo_url: https://github.com/whatever/whatever.git
  overleaf1:
    kind: overleaf
    project_id: skdjrheohrwejhrwe
  gdrive:
    kind: google-drive
    folder_id: edkjhew98fekjhdf

# By default, paths go into the main git and dvc storage, which are implicit based on the project owner/name
paths:
  "path/to/thing": 
    storage: git1
    external_path: "thing"
    sync: bidirectional
    external_rev: dfdkj5445dk
```

We could then not commit them to our repo perhaps, and then push/pull operate only on the external storage providers?

Another use case is a `references.bib` file that live in its own repo. Instead of adding a submodule, we could import and sync that file, optionally bidirectionally, such that references added in child projects end up making their way back into the main collection.

Another use case is CliMA config files. There's one in ClimaCoupler that's useful for developing ClimaAtmos. It might be nice to be able to pull in changes to it that apply on top of the changes we've made locally.

petebachant:
Maybe relevant: https://github.com/open-condo-software/gitexporter

Issue #264, opened 2025-02-19
I want to be able to track and update files like scripts and simulation configs imported from other projects


petebachant:

petebachant:
The Calkit dev container spec is an example here. Could we build this "imported path" concept into the CLI, or perhaps it should be another tool for this general concept of sharing assets between repos (Git, Zenodo, URLs, etc.)? We could call it Vandelay after Costanza's fictitious importer/exporter persona on Seinfeld. The scope would be: repos declaring exported paths for others to import, and repos importing from elsewhere with a tool to check for updates, pull them in, potentially merge/bidirectionally sync like we do for Overleaf.

petebachant:
PR #1579 implements the core of this, and arrived independently at the design sketched in *Another design* — including the filename:

- `calkit import path <src> [dest] --kind <kind>` records where a file came from, from a Git repo (https/SSH clone URLs, forge file links), a URL, a Calkit project, or a DOI.
- `calkit update import <path>` / `--all` refreshes them; `calkit list imports` lists them across every artifact kind.
- Resolved state — the commit, a sha256, when it was fetched — lives in **`.calkit/imports.json`**, keyed by path and committed. `calkit.yaml` keeps only what a person declared (`repo_url`, `path`, `ref`), so pinning is writing a commit hash as the `ref`.

On the first open question — *"How to track differences, e.g., if a file was modified after import?"* — that checksum is the answer, and the behavior matches the `pull-only` sketch exactly:

```
Error: 'scripts/setup.sh' has been edited since it was imported, and refreshing it
would discard that; pass --force to overwrite, or drop its 'imported_from' if it is
now maintained here
```

**Not** covered, so this issue shouldn't be closed on that PR alone: a top-level `imports:` section with per-import `sync` config; submodules, subtrees, and Overleaf repos as import kinds; `derived_from` for versions that intentionally diverge; releasing scripts to Zenodo for citation; and the "breadcrumbs in projects we push a synced path to" idea.

The imports-vs-syncs naming question at the end came up during review too — an import is an inbound-only sync, which is why its record is a *lock* (committed, shared) rather than a *merge base* (local, per-checkout, like `.calkit/overleaf-sync.json`).

*(From Claude, prompted by @petebachant.)*

PR #1579, opened 2026-08-26
Add setup to stages and system envs and misc path import/sync

### LINKED ISSUE #78: I want to be able to update an imported object to the latest version
## Potential design

```sh
calkit update-imported --all
```

should go through all of our imported objects and update them to the latest version.

```sh
calkit update-imported figure path/to/the/figure.pdf
```

Should commit by default, probably.

## Alternatives

Does this need to go through the API? If objects were imported from Git repos that aren't setup as Calkit projects, or just plain URLs, should we support that?

petebachant:
Resolves #78. Addresses part of #264. Related to #1339, #1331, #728.

## TODO

- [ ] Provide a more explicit setup mechanism instead of simply a command string?

```yaml
environments:
  compilation:
    kind: system
    requirements:
      - ...
    lock: # What properties of the system should be tracked and invalidate outputs?
      - machine-id
      - os-version
      - cpu-count
    default_setup:
      - kind: source
        path: scripts/setup.sh
```

The above would make the discipline of setting `inputs` unnecessary. We could also detect the inputs based on the command, and potentially reject those outside the project dir, like we do for stage path inputs (do we do that for all? We should)

- [x] Git imports can track a ref, if specified, but will always need a rev too. Ref is optional, e.g., if someone wants to always update from a branch?
- [ ] `.calkit/imports.json` should simply be `.calkit/syncs.json`, where imports are tracked as pull-only syncs? Sort of setting us up for the full #728.
- [ ] Inline environments into `.calkit/inline-envs/{stage_name}`?
- [ ] `misc` to `inputs` in calkit.yaml? We could have datasets that are inputs though?


calkit/pipeline.py
```diff
@@ -1494,7 +1494,7 @@ def to_dvc(
     """
     import calkit.dvc.zip
     import calkit.markdown
-    from calkit.environments import get_env_lock_fpath
+    from calkit.environments import env_input_paths, get_env_lock_fpath
```
petebachant: Function names usually start with a verb

```suggestion
    from calkit.environments import get_env_input_paths, get_env_lock_fpath
```

petebachant:
**From Claude** (Anthropic's Claude Code), prompted by @petebachant, from a real case in the boundary-layer-turbulence-modeling project. The prompt, quoted:

> seeing some irrationality here in the fetching of the jhtdb gradient fetching. That should be a pipeline stage, but perhaps needs a requirements declaration for the JHTDB access token. Will require PR 1579 so we can define requirements for the jhtdb environment? Would need to be nested--a uv env inside a system env with the JHTDB access token inside. may need a new kind of requirement to express either a file or an env var.

## The case

`scripts/standalone/fetch-jhtdb-gradients.py` queries the JHTDB web service for exact velocity gradients at sampled points. It is deliberately *not* a pipeline stage today, for two reasons that are both really gaps in what `calkit.yaml` can say:

1. It needs a credential: a JHTDB access token, which `pyJHTDB` reads from `~/.config/JHTDB/auth_token.txt` (an env var would do equally well).
2. It runs in its own `uv` environment (`py-jhtdb`, pinned to Python 3.11 / numpy < 1.24 because `pyJHTDB` builds from source against removed APIs), so the requirement belongs to the *environment the stage runs in*, but that environment is not a `system` one and today only `system` environments and the project as a whole carry `requirements`.

The consequence is that a token-gated download lives outside the pipeline, its output (`data/jhtdb-transitional-bl/sampled-gradients.h5`) is declared as an imported dataset rather than a stage output, and the provenance trail stops at "someone ran a script".

## What this PR already gives, as I read it

- `requirements` on the project and on `system` environments, with kinds `app`, `env-var`, `setup` (check/setup commands, cached) and `calkit-config`.
- `default_setup` on system environments: commands run in the stage's shell, recorded in the lock, with declared `inputs`.

With only that, the closest expression is a project-level `setup` requirement whose `check_command` is `test -f ~/.config/JHTDB/auth_token.txt`. It works, but it gates the *whole project* on a credential that one stage needs, and it can't say "or the `JHTDB_TOKEN` variable".

## Requirements Calkit would need for this to be a stage

1. **A `file` requirement kind** — a path that must exist, with `~` expansion, optionally with a `description` telling the user how to obtain it. The token file is the common case for scientific web services (JHTDB, NASA Earthdata `~/.netrc`, `~/.config/gh`, kaggle.json).
2. **Alternatives** — a requirement satisfied by any one of several: `any_of: [{kind: env-var, name: JHTDB_TOKEN}, {kind: file, name: ~/.config/JHTDB/auth_token.txt}]`. Credentials are almost always accepted in more than one form, and forcing one form into `calkit.yaml` makes the project wrong on half the machines.
3. **Requirements on non-system environments and on stages.** The token is a property of *where `pyJHTDB` runs*, i.e., the `uv` environment `py-jhtdb`, or of the single stage that calls it. Either `requirements:` on every environment kind, or `requirements:` on a stage, so that `calkit run` on a machine without the token still runs everything else.
4. **Requirement-gated stages, not failed runs.** When a stage's (or its environment's) requirements are unmet, `calkit run` should *skip it and keep its committed outputs*, reporting "cannot run here: missing `JHTDB_TOKEN` or `~/.config/JHTDB/auth_token.txt`", and `calkit status` should show it as gated rather than stale. That is the difference between a pipeline that is reproducible-with-a-caveat on a fresh clone and one that is red. It also matters for the hub and CI, which will never hold the token.
5. **Nesting: a `uv` environment inside a `system` environment.** The system environment is where the credential lives and where `default_setup` could export it (`export JHTDB_TOKEN=$(cat ~/.config/JHTDB/auth_token.txt)`); the `uv` environment is the interpreter the stage needs. Today an environment is one or the other. Something like `environments.py-jhtdb.within: jhtdb-host` (or `stage.host_environment`) so that the system env's requirements, `default_setup` and lock apply, and the uv env is created and used inside that shell. The same nesting is what an HPC user needs for "uv env inside a SLURM environment with modules loaded", so it is not JHTDB-specific.
6. **Secrets stay out of the record.** `env_vars` on Docker environments record values; a credential must never land in `.calkit/env-locks`, `dvc.lock`, or a log. A `secret: true` flag on an `env-var` requirement (name recorded, value never) would make that explicit, and the provenance of the fetched data should record *that* a token was required, not which one.

## What we would write, given 1–5

```yaml
environments:
  jhtdb-host:
    kind: system
    requirements:
      - any_of:
          - {kind: env-var, name: JHTDB_TOKEN, secret: true}
          - {kind: file, name: ~/.config/JHTDB/auth_token.txt,
             description: "Request a token at turbulence.pha.jhu.edu"}
  py-jhtdb:
    kind: uv
    path: envs/jhtdb/pyproject.toml
    within: jhtdb-host

pipeline:
  stages:
    fetch-jhtdb-gradients:
      kind: python-script
      script_path: scripts/fetch-jhtdb-gradients.py
      environment: py-jhtdb
      outputs:
        - data/jhtdb-transitional-bl/sampled-gradients.h5
```

and the dataset entry would drop its `imported_from` in favor of `stage: fetch-jhtdb-gradients`, with the token requirement visible in the provenance.

Not asking for all of it in this PR; items 3 and 4 are the ones that turn a credential-gated script into a stage, and 1–2 are what make the requirement honest.

petebachant:
**From Claude**, a follow-up prompted by @petebachant:

> for the closure challenge data. It says we imported from an arxiv record, but is that actually where the data was fetched from? Again, if a direct download of that is not possible, make it a pipeline stage. That should be part of Calkit's reproducibility check--downloads happen in the pipeline if possible, so we can point to a stage in the dataset definition

One more requirement, on the provenance side rather than the credential side:

7. **Downloads belong in the pipeline, and the check should say so.** A dataset that was fetched from a URL or a Git repository, but whose entry names no `stage`, is a download someone did by hand once. `calkit check` (or the reproducibility check) should flag any dataset with an `imported_from` of the `url` or `git` form and no `stage`, with the fix being "make the fetch a stage and point `stage:` at it", and it should accept `stage` *together with* `imported_from` — the stage says how it was fetched, `imported_from` says where from. Today the model allows both, which is right; the check that asks for both is what's missing. The exception is exactly the credential-gated case above: a stage that cannot run without a token is still a stage, and items 3–4 are what let it be declared without breaking every other machine's run.

In the project this came from, the Closure Challenge cases are now recorded as `imported_from: {git: {repo_url, rev, path: data}}` at the pinned commit *and* `stage: fetch-dns-data`, which is the form 7 would require. The `doi` form is reserved for the paper being cited, which is a different thing from where the bytes came from.

calkit/cli/main/core.py
```diff
@@ -2942,6 +2943,29 @@ def run_in_env(
...
+        typer.Option(
+            "--env-default-setup",
+            help=(
+                "How to apply the environment's default setup commands: "
+                "'replace' (default) uses them only when no setup commands "
+                "were given here; 'merge' runs them first; 'ignore' never "
+                "runs them."
+            ),
+        ),
+    ] = "replace",
```
petebachant: Not sure if these should be in the CLI or simply in a stage's compiled command string directly. That is, this logic may be better inside pipeline compilation. Worth a thought.

calkit/cli/main/core.py
```diff
@@ -3025,7 +3097,7 @@ def save_env_check_cache() -> None:
                     env=env, env_name=env_name, as_posix=False
                 ),
                 platform=env.get("platform"),
-                deps=env.get("deps", []),
+                deps=calkit.environments.env_input_paths(env),
```
petebachant: Functions should start with verbs:

```suggestion
                deps=calkit.environments.get_env_input_paths(env),
```

calkit/cli/import_.py
```diff
@@ -227,6 +227,164 @@ def import_dataset(
...
+        str,
+        typer.Option(
+            "--kind",
+            help=(
+                "Which list in calkit.yaml to record this in: 'datasets', "
+                "'figures', 'publications', or 'misc' (default), which is "
+                "where a path that isn't one of the typed artifacts belongs."
+            ),
+        ),
+    ] = "misc",
```
petebachant, line 264: We're sort of locked into a unique "table" for each artifact schema here, where we may have been able to get away with a single table for all artifacts with a `kind` key, sort of like the ASTRA spec's inputs/outputs. There are pros/cons to each approach, e.g., certain types of artifacts may have properties others don't. But YAML is more of a document database than a relational one, so maybe the single list for all is smarter. Perhaps we could support both at some point, but no need to diverge now.

calkit/cli/new.py
```diff
@@ -1153,8 +1154,8 @@ def new_docker_env(
             env["env_vars"][key] = value
     if args:
         env["args"] = args  # type: ignore
-    if deps:
-        env["deps"] = deps  # type: ignore
```
petebachant, line 1189: We'll keep reading from both, but display a deprecation warning for deps, right?

calkit/tests/cli/main/test_core.py
```diff
@@ -585,6 +585,114 @@ def test_run_in_env_system(tmp_dir):
...
+    with open("calkit.yaml", "w") as f:
+        f.write(
+            "environments:\n"
+            "  sys2:\n"
+            "    kind: system\n"
+            "    default_setup:\n"
+            "      - source setup_env.sh\n"
+            "    deps:\n"
+            "      - setup_env.sh\n"
+        )
```
petebachant: Why not dump a dict as YAML?

calkit/tests/cli/main/test_core.py
```diff
@@ -585,6 +585,114 @@ def test_run_in_env_system(tmp_dir):
...
+    # only when the stage names none, 'merge' runs the env's first, and
+    # 'ignore' leaves them out
+    with open("calkit.yaml", "w") as f:
+        f.write(
+            "environments:\n"
+            "  sys3:\n"
+            "    kind: system\n"
+            "    default_setup:\n"
+            "      - export CK_TEST_ORDER=env\n"
+        )
```
petebachant: Again, weird to not use YAML

calkit/tests/cli/main/test_core.py
```diff
@@ -585,6 +585,114 @@ def test_run_in_env_system(tmp_dir):
...
+    # A failing setup command stops the stage rather than running it
+    # without whatever the setup was meant to provide
+    with open("calkit.yaml", "w") as f:
+        f.write(
+            "environments:\n"
+            "  bad:\n"
+            "    kind: system\n"
+            "    default_setup:\n"
+            "      - exit 3\n"
+        )
```
petebachant: Again

calkit/tests/cli/test_import.py
```diff
@@ -23,3 +23,240 @@ def test_import_zenodo(tmp_dir, monkeypatch):
...
+    import os
+
+    import calkit
+
+    def git(*args, wdir):
+        subprocess.run(
+            ["git", "-C", wdir, *args],
+            check=True,
+            capture_output=True,
+        )
```
petebachant: We do have GitPython at our disposal here...

docs/provenance.md
```diff
@@ -64,6 +64,97 @@ misc:
 Scripts count. A figure resting on a script nobody will claim has just as
 little behind it as one resting on data nobody collected.
 
+### Bringing a file in from elsewhere
+
+`calkit import path` copies a file into the project and writes the entry
```
petebachant, line 69: Should it be `calkit import misc` since it goes into the `misc` list in calkit.yaml? Then `calkit update misc`? That's a little harder on the user since they need to know the category. So it seems to make sense to keep `calkit update path` and `calkit import path` with a `--kind` option.

docs/provenance.md
```diff
@@ -64,6 +64,97 @@ misc:
...
+
+A `ref` naming a commit rather than a branch is a fixed point, so an
+import pinned that way---including one taken from a link to a file at a
+particular commit---stays where it is when refreshed.
+
+Either way this is a one-way copy from the source, not a merge: local
+changes to the file are discarded. An import is a claim about where the
+bytes came from, and an edit that survived would make that claim false.
+If the file needs to differ here, it isn't imported---drop the
+`imported_from` and record who changed it instead.
```
petebachant, line 165: We should have a `calkit list imports` function if we don't already. This should go through all artifact categories and `misc`, printing all imported artifacts.

calkit/tests/cli/main/test_core.py
```diff
@@ -572,27 +572,148 @@ def test_run_in_env_detect_default(tmp_dir):
 
 
 def test_run_in_env_system(tmp_dir):
+    def _write_ck_info(environments: dict) -> None:
```
petebachant:
```suggestion
    def _write_envs_to_ck_info(environments: dict) -> None:
```

docs/provenance.md
```diff
@@ -64,6 +64,164 @@ misc:
...
+`calkit.yaml` records what a person declared---the repo, the path within
+it, and the `ref` to follow. What a fetch resolved to goes in
+`.calkit/imports.json`, keyed by path:
+
+```json
+{
+  "scripts/setup.sh": {
+    "fetched": "2026-08-31T23:14:21+00:00",
+    "rev": "4fadbcf62125c19c9cbf31de60831f656ffe5d4e",
+    "sha256": "sha256:9f2b..."
```
petebachant: Should this just be `hash` since the algorithm is part of the value?

docs/provenance.md
```diff
@@ -64,6 +64,164 @@ misc:
...
+To pin an import, write the commit hash as its `ref`. A commit is a thing
+to follow that happens never to move, so no separate field is needed and
+`calkit.yaml` stays a statement of intent.
+
+The recorded checksum is what makes a local edit visible. Refreshing a
+file that has been changed since it was fetched would discard that work,
+so it is reported and refused until `--force` says otherwise:
+
+```sh
+calkit update import scripts/setup.sh
```
petebachant: This could also be `calkit sync import`.

Issue #102, opened 2024-11-22
Add unit test for `run_procedure`


petebachant:
Can use something like this to simulate input to the CLI:

```python
import pytest

def get_user_data():
    name = input("Enter your name: ")
    age = int(input("Enter your age: "))
    return name, age

def test_get_user_data(monkeypatch):
    # Simulate user input
    inputs = ["John", "25"]
    monkeypatch.setattr('builtins.input', lambda _: inputs.pop(0))

    # Test the function
    name, age = get_user_data()
    assert name == "John"
    assert age == 25
```

Issue #989, opened 2026-06-24
I want to be guided towards the next set of parameters in an expensive manual experiment


petebachant:
https://www.youtube.com/watch?v=Wyeab_JNBAo

Issue #305, opened 2025-04-01
I want to be able to run a parameter sweep in parallel across many machines


petebachant:
Inspired by https://www.reddit.com/r/OpenFOAM/comments/1jp25yb/easiest_cloud_solution_to_batch_runs/

>I've got a simulation that runs in about 10 minutes on my desktop and I'd like to rerun with 1000-10000 different geometries (exactly the same case except different blockMeshDict). What's the easiest and fastest way to get this running on the cloud for someone with no experience with AWS or similar? Has anyone used AWS Batch or Lambda?

From a local machine, I'd like to define a massive pipeline and run these remotely on a collection of machines. Each should either send its results back or commit and push so I can pull them back.

Being able to quickly spin up the environment on the remote machine would be nice. Docker or Apptainer can help there.

petebachant:
Also relevant https://www.reddit.com/r/Jupyter/comments/1m6704p/any_method_to_offload_compute_to_a_local_network/

Issue #12, opened 2024-08-24
I can easily save data from an experiment or device


petebachant:
```python
calkit.write_data(df, "local/path/somefile.parquet", commit=True, push=False)
```

This will update our `data.yaml` file if necessary, do some DVC stuff, some Git stuff, and optionally push to the cloud.

Should have the ability to write a file and perhaps go into a dataset that is a folder. Maybe an optional `dataset_path` parameter.

This would be great if we could save to an Iceberg table, maybe with the metadata saved locally. Someday we can put these into catalogs.

Or, we could have devices not clone the repo, but instead push data to our API, and we do the Git/DVC stuff.

PR #1600, opened 2026-08-29
Inject project content into documents with provenance

petebachant:
Inject project content into documents with provenance, and resolve what a document takes from the project.

Builds on #1619, which carries the question checking underneath this. Rebase once that lands.

- Values and questions injected into LaTeX, `calkit.sty` to mark them, and a provenance record named after its artifact with a published schema
- `calkit describe components`: every value and figure a document takes from the project, where each is written, the stage behind it, and whether it is current
- The same lens in the VS Code extension, on Overleaf via the browser extension, and in the hub's publication components API and UI

## TODO

- [ ] Compare with `showyourwork`


petebachant:
**From Claude** (Anthropic's Claude Code), prompted by @petebachant. The prompt, quoted:

> I don't think we need the github hyperlinks in the pdf for now, nor do we need the figure path and stage showing up there. The VS Code extension and web app (hub) should link them within those systems, i.e., if I hover over a figure or value I can open up the relevant file, and from there I can see its stage, etc. The optimal workflow is to be able to quickly open the script for whatever produces it to make a quick iteration, then hop back into writing with minimal context switching.

## Plan: following the provenance trail from the editor and the hub

The branch now leaves the PDF alone apart from coloring injected values (`\usepackage[provenance]{calkit}`), and puts the trail in one machine-readable place per document, `<document>.provenance.json`: every value, figure and text block the document took from the project, with the pages it appears on, the producing stage, that stage's inputs, and the hash from `dvc.lock`. Everything below reads that file plus `calkit.yaml`; nothing new is needed in the document.

### 1. Resolving what is under the cursor

A small resolver in `calkit` (`calkit describe injection <document> --line N --col M`, JSON out; also importable) maps a position in the TeX source to a provenance record:

- `\result[Key]` / any command generated by `json-to-latex` → the results file and key, via the `\ckvalue` expansion recorded in the generated `.tex`;
- `\ckfigure{...}`, `\includegraphics{...}` → the figure path;
- `\ckanswer[n]` / `\ckfindings` → question `n` in `calkit.yaml` and each value evidence entry under it;
- from any of those, the stage that produced the path (`get_stage_for_output`), the stage's script or notebook (`script_path` / `notebook_path`), and its inputs.

The same resolver works from the PDF side for viewers that support SyncTeX: a click in the PDF gives a source line, and the source line gives the record. The page list in the sidecar is the fallback when there is no SyncTeX.

### 2. VS Code extension

- **Hover** on `\result[Key]`, `\ckfigure{...}`, `\ckanswer[n]`: the current value (or a thumbnail for a figure), the source path, the stage, and links: *Open file*, *Open script*, *Open in hub*.
- **Go to definition** (F12) on a value or figure jumps to the results file at the key (JSON/YAML) or opens the figure; **Go to declaration** opens the producing stage's script. That is the "open the script, tweak, come back" loop in two keystrokes, with the editor's own back-navigation returning to the paragraph.
- **CodeLens** above a figure environment or an appendix block: `stage plot · 3 inputs · run stage`, where *run stage* executes `calkit run -s <stage>` in the integrated terminal and the document rebuilds through the pipeline as usual.
- **Diagnostics** from `calkit check questions --json` and the sidecar: a stale question, a placeholder that names no evidence, a `\result` key that no longer exists in its file, a figure the pipeline no longer produces.
- Implementation: a language-server-style provider in `vscode-ext` calling the resolver over the CLI (JSON), no TeX parsing in TypeScript beyond finding the command under the cursor; `calkit.yaml` and the generated `.tex` are watched so hovers stay current after a run.

### 3. Hub (web app)

- The publication page reads `<document>.provenance.json` and shows an *injections* panel: every value, figure and block, grouped by page and by stage, each linking to the artifact page, which already shows the stage, its inputs and the lineage graph.
- A *reverse* view on the artifact page: "used in `paper/main.pdf` on pages 3, 7", so a change to a result shows what it touches.
- The PDF viewer overlays the page list from the sidecar: highlight the injected values on the current page; click one to open its artifact. This needs no links in the PDF itself and works for any format that writes the same sidecar.
- Questions appear on the project page rendered from evidence (the same rendering `calkit list questions` does), with staleness from `calkit check questions`.

### 4. Other formats

The sidecar is the contract. Markdown/MyST/Quarto get a preprocessing stage that resolves `{{ ck:value path:key }}` and writes the same record; DOCX via content controls bound to a custom XML part; PPTX via alt-text. The resolver and both front ends are unchanged.

### Order

1. Resolver + `calkit describe injection` (small, testable, shared by both front ends).
2. VS Code hover and go-to-definition/declaration, then CodeLens *run stage*.
3. Hub injections panel and reverse view.
4. Diagnostics from `check questions` in the editor.

Not in scope until asked: links inside the PDF (removed in this branch), and any marker text on the page beyond color.

petebachant:
One important aspect of inspecting the paper PDF is being able to know when one of the components (figures, tables, values) is stale per the pipeline status, indicating we need a rerun.

This is also true for the browser extension when the user is working on Overleaf. We should be able to see where components came from, if they're up-to-date, and quickly jump into a mode to edit them.

petebachant:
**From Claude** (Anthropic's Claude Code), prompted by @petebachant.

## TODO list: what's done

### 1. Questions in status is slow

Both fixed, since they were different problems.

**It was slow because of an O(questions x commits) walk.** Finding the commit a
question was last edited at walks `calkit.yaml`'s history until the question
differs, and every question walked it independently: with Q questions and C
commits that's Q x C `git show` subprocesses plus Q x C round-trip YAML parses,
to answer C commits' worth of history. Three changes:

- one `CalkitYamlHistory` shared across questions, so each revision is read once;
- blobs read from the object database instead of shelling out to `git show` per
  commit (20x cheaper for the same bytes);
- PyYAML rather than the round-trip parser for history, which is never written
  back.

Measured on 8 questions over 61 commits of `calkit.yaml`: **13.8s -> 0.66s**. A
regression test asserts each revision is parsed once rather than once per
question.

**It's also out of the default status categories now**, per your suggestion.
`calkit status` no longer runs it; `calkit status -c questions` does, and
`calkit check questions` remains where it belongs. Judging whether an answer
still matches its evidence means reading `calkit.yaml`'s history, which nothing
else in `calkit status` needs, and the report is a page of its own once a
project has a few questions.

### 2. Working backwards from the writing

Three parts.

**A `procedure` pipeline stage kind.** Procedures already existed and could
already join a pipeline, but only as a hand-written raw DVC block --- you had to
know to write `cmd: calkit xproc <name>` with `cache: false` and `persist: true`
on the log directory. Now:

```yaml
pipeline:
  stages:
    collect-data:
      kind: procedure
      procedure_name: measure-rig
    plot-data:
      kind: python-script
      environment: py
      script_path: scripts/plot-data.py
      inputs:
        - from_stage_outputs: collect-data
      outputs: [figures/temperature.png]
```

`calkit run` walks the person through the steps, and the run log is an output
the next stage reads like any other. The log directory is declared for you, in
Git and never cleared before a run (earlier runs are data, not stale output).
The stage depends on `calkit.yaml`, and on the procedure's own file when it's
kept in one, so editing the steps means the procedure needs carrying out again
--- verified end to end: done once it's skipped on rerun, edited and committed
it asks again, and the earlier log survives.

This needed one behavior change worth flagging: **`calkit xproc` now requires
only the procedure's own definition to be committed, not the whole working
tree.** The old whole-tree check could never hold for a stage, since earlier
stages have written their outputs by the time one runs. The narrower check keeps
the guarantee that actually mattered ("you ran the committed version of the
procedure") and drops strictness that said nothing about which procedure was
carried out. The per-step commit already only stages the log file, so unrelated
work in progress was never swept up.

**Closes #102** --- `run_procedure` had no unit test, and I was changing it. Ten
now, using the `monkeypatch` on `builtins.input` approach from that issue:
inline and file-referenced procedures, input coercion and retry on a bad value,
the per-step commits and `--no-commit`, the narrowed definition check, and the
stage's compilation.

**A writing-first workflow.** A `/calkit:answer-todos` agent skill and a
[tutorial](docs/tutorials/writing-first.md): you write the paper with holes in
it and mark them, and the agent builds the pipeline that fills them. The rule
the skill works under is that it never types a computed number into the
manuscript --- a number typed into a `.tex` file is one nobody can check, and it
goes stale the moment a stage reruns, so everything reaches the page through
`json-to-latex` and `\result[...]`. A TODO it can't close because the data
doesn't exist becomes a `procedure` stage rather than a plausible-looking
number, which is the case the stage kind above exists for.

## Future work, not implemented

Complex experimental data collection --- psychology studies and similar --- is
bigger than this PR, and most of it already has issues open. What a procedure
can't do today, mapped onto them:

- **Multiple participants or sessions.** A procedure is one linear run by one
  person. A study is N participants x M conditions, and the log is one CSV per
  run with no notion of who or which condition. Randomization, counterbalancing
  and blinding have nowhere to live. Related: **#305** (parameter sweeps / DoE),
  though that's the compute-side version.
- **Adaptive designs.** Steps are fixed ahead of time; nothing branches on what
  was just entered, or on a fit over what's been collected so far. **#989** is
  exactly this ("guided towards the next set of parameters in an expensive
  manual experiment"), and there's a `# TODO: Add ability to process periodic
  logic` in `run_procedure` for the scheduling half.
- **Anything but a terminal.** Reading a prompt off a laptop is fine at a bench
  and wrong for a participant. **#1268** covers running a procedure from the web
  app, including on mobile and with images or diagrams per step, which is most
  of what a participant-facing runner needs.
- **Surveys.** **#171** already asks for survey data going straight into a
  dataset. Worth noting that a survey and a procedure want nearly the same
  model: an ordered set of prompts with typed inputs, a per-response log, and a
  results file downstream stages read.
- **Instrument capture.** **#12** ("save data from an experiment or device").
  Procedure inputs are typed by a person; a lot of collection is a device
  writing a file that should land as a stage output.
- **Ethics and consent.** Nothing records IRB approval, a consent version, or a
  participant identifier that isn't PII. This one has no issue that I can find,
  and it's the one that would block real human-subjects use.

The shape that would cover most of it: a procedure that iterates like
`iterate_over` does (over participants or conditions), writes one log per
iteration keyed by an identifier, and can be carried out through something other
than a terminal. That's a design conversation rather than a change, so I've left
it here rather than opening an issue that guesses at it.

Nothing above is in this branch. Say the word if any of it should be an issue of
its own rather than a comment.

petebachant:
**From Claude** (Anthropic's Claude Code), in discussion with @petebachant.

The observation that started this: `build-paper-pipeline` (the writing-first skill on #1619) is adjacent to literate programming --- runnable READMEs, notebooks, Quarto, MyST --- and that family has always assumed everything lives in the document. That assumption is unmodular. A script that runs a few hundred CFD cases does not want to be configured and launched from inside a paper. But a document that is *connected* to a pipeline stage, bidirectionally, so the stage can be opened and edited in its own context without hand-editing the document, is a different proposition.

## Containment vs. citation

The distinction worth naming is whether the document **contains** the code or **cites** it.

Notebooks, Quarto, MyST and runnable READMEs contain it. The document becomes the unit of editing, running, diffing, reviewing and reuse, which is why a large analysis never fits: you cannot run it headlessly without the document, cannot test it, cannot reuse it from a second document, and cannot read its diff without reading the prose around it.

`\result[Key]` cites it. The document names an artifact the pipeline owns, and the stage keeps its own file, its own environment, its own inputs and outputs, its own run. Nothing about the analysis has to fit in a paragraph. This is literate programming by reference, and it is the only version of it that survives a real analysis --- the paper stays a paper, and the CFD driver stays a program.

## Navigability is what makes citation work

Containment has one genuine advantage: everything is right there. A citation you cannot follow is strictly worse --- you lose that, and get nothing back. Make it followable in both directions and the trade reverses:

- from a number in the manuscript, open the stage that produced it, edit, rerun, and come back --- without switching context or hand-syncing anything;
- from a stage, see which documents cite it, and which of their values and figures go stale when it reruns.

That is what this PR is building, and it is the part that turns "the document cites the pipeline" from a constraint into an affordance. It is worth being explicit that this is the point of the components lens, rather than treating it as a convenience feature bolted onto the injection work.

## The sidecar is the portable contract

The reason this generalizes past LaTeX: what a viewer needs is not the macro, it is the record. `<document>.provenance.json` says what the document took from the project, where each piece came from, what stage produced it, and whether it is current. The format-specific part is only how a spot in the document is marked and located:

| Format | Marker | Position |
| ------ | ------ | -------- |
| LaTeX | `calkit.sty` macros | source line, or SyncTeX from the PDF |
| DOCX | content controls bound to a custom XML part | control id |
| PPTX | alt-text on the shape | slide and shape id |
| Markdown / MyST / Quarto | a directive resolved by a pre-processing stage | source line |

Everything above the marker --- the resolver, the editor lens, the hub panel, the staleness check --- reads the sidecar and does not care which one it was. An Office add-in is then the same shape of work as the VS Code extension already in this PR, not a second system.

## Calkit already does both, and that is fine

Markdown stages (the runnable README) are containment: a fenced block becomes a stage. That is the right call for glue whose only reason to exist is the document --- a plot tweak, a small table, three lines of setup. Extracting it to a file would be ceremony.

So the rule is not "citation always". It is:

> **Containment for code with no life outside the document. Citation for anything with a life of its own.**

And the interesting move that falls out of a bidirectional link: promoting a block from one to the other becomes a refactoring rather than a rewrite. Extract a markdown block into a real stage, and the document keeps citing it under the same name --- the reference does not change, only what is behind it. That is the escape hatch literate programming has always been missing: the small thing that grew is not trapped in the page it was born on.

## Where this leaves the two PRs

Nothing here changes what is in this branch; it is a frame for what the branch is already doing, and an argument for which parts to keep pushing. Worth noting that the writing-first skill on #1619 is already on the citation side by construction --- a TODO comment is a brief, not code, and the skill's hard rule is that no computed number is ever typed into the manuscript.

If this is worth pursuing past this PR, the pieces that follow from it are: a position-to-record resolver as the shared primitive (the editor, the hub and the PDF viewer all want it), the DOCX and PPTX markers, and "extract this block to a stage" as an actual command. Happy to open those as separate issues rather than let them expand this one.

PR #1631, opened 2026-09-02
Improve repo cache warming with shared read-only clones

petebachant:
## TODO

- [x] Fix first figures page load latency
- [x] Create a pre-commit hook in the read-only clones to error if a commit is attempted
- [x] Make absolute sure read-only clones have no GitHub write access credentials left over in them that could allow unauthorized access

### DIFF
--- a/hub/backend/app/git.py
+++ b/hub/backend/app/git.py
@@ -181,6 +181,61 @@ def get_remote_head_sha(
     return sha
 
 
+# Where the one checkout everybody reads from lives, under CLONE_ROOT. Not a
+# possible account name (see INVALID_ACCOUNT_NAMES), and the leading
+# underscore keeps it from colliding with one anyway.
+SHARED_READER_DIR = "_shared"
+
+
+def is_shared_read_checkout(repo: git.Repo) -> bool:
+    """Whether this repo is the checkout everybody reads from."""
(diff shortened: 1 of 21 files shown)


hub/backend/app/git.py
```diff
@@ -181,24 +181,98 @@ def get_remote_head_sha(
...
+
+
 def get_repo(
     project: Project,
     user: User | None,
     session: Session,
     ttl: int | None = None,
     fresh=False,
     ref: str | None = None,
+    shared_read: bool = False,
```
petebachant: This could probably be simply `read_only`, and the fact that it's shared between callers is more of an internal detail to the Git subsystem.

hub/backend/app/git.py
```diff
@@ -181,27 +182,142 @@ def get_remote_head_sha(
     return sha
 
 
+# Where the one checkout everybody reads from lives, under CLONE_ROOT. No
+# GitHub name can be spelled this way, and ``_clone_dir_segment`` keeps the
+# per-user directories out of it too.
+SHARED_READER_DIR = "_shared"
```
petebachant, line 189: Is this in the protected accounts list? Would this conflict with write-access path for an account?

petebachant: We should probably not allow underscores in account names, like GitHub does, anyway.

PR #1647, opened 2026-09-06
Add local LaTeX to Word to LaTeX collaboration/review workflow

### LINKED ISSUE #1529: My collaborators want to be able to mark up a document in Word but I want it to stay LaTeX
What we need is some sort of round-trippable conversion: TeX to Word, Word to TeX, TeX to PDF, all properly part of the pipeline so we know when the final PDF is out-of-date.

The most faithful Word doc from PDF may come from opening in Word itself.

This could be part of a review request workflow (#1485). Generate the docx, email, wait for it to be re-uploaded, convert back, merge, and keep working in TeX (or Markdown, or typst, etc.)

Related: https://www.reddit.com/r/Markdown/comments/1vshe4u/i_built_a_compiler_for_word_documents_because_i/

Also: https://www.reddit.com/r/LaTeX/comments/1w4jvlr/the_best_way_to_convert_a_tex_to_a_word_file/

## Design: `calkit latex docx-sync`

We have a LaTeX Word sync session that starts and its state saved in the repo. We keep track of paper PDF hash, all inputs, etc.

PDF is compiled and converted to Word to start the session. The docx is ready to email and has special metadata so we know that it will be okay to try to merge back into the TeX source.

After collaborators make their edits and comments, we ingest them into the current session. This is an interactive process that allows us to cherry-pick their edits. Comments make their way into a separate file with our responses. Might be nice for it to go into git-bug someday, but baby steps here.

petebachant:
Resolves #1529 

The design here is intentionally low coupling. You can do everything from the CLI and most of the important data lives inside the docx. The user has some tracking to do on their own, so we write some data out to `.calkit/latex` to help, but we don't auto-commit any of this, since `calkit latex` is lower level. A future `calkit review` or `calkit contrib` command might control a more integrated form of this workflow.

## TODO

- [x] Remove monospace formatting from file extensions in docs.
- [ ] Simplify Claude's code.
- [x] Tutorial should show the `uvx ck9` alias, or at least have it as a tip?
- [x] Warn if a figure is edited in Word.
- [x] Test a table round trip.
- [x] Manual test on Windows with a real Word install to check mocks.
- [ ] `to-word` instead of `to-docx`?
- [ ] Should be able to point at .tex or .pdf file when converting to Word?
- [x] Rationalize record paths. Maybe should be by the path being operated on, not UUID. Records should probably also use `id` as the key, not `uuid`.
- [ ] Record to `.calkit/local` by default?
- [x] Add `ck new project --from https://github.com/calkit/calkit/examples/latex-word` and more commands so users can try it out.
- [x] Document LaTeX comment schema on LaTeX page.
- [x] Manual testing in non-calkit projects.
- [x] Ensure sync records have paths and hashes of files involved at sync time.
- [x] Comment timestamps?
- [ ] Merge needs to be done from the same wdir as the export--may want to make that more robust


calkit/cli/latex.py
```diff
@@ -613,3 +613,426 @@ def _build_diff(
...
+        dirty = any(repo.is_dirty(path=p) for p in {ln.path for ln in lines})
+    except Exception:
+        pass
+    export_id = str(uuid.uuid4())
+    doc.write_original(
+        calkit.docx.Original(
+            export_id, rev, source, original, doc.media_hashes()
+        )
+    )
+    doc.set_identifier(f"calkit-review:{export_id}:{rev or ''}:{source}")
```
petebachant: This doesn't necessarily have to be used for reviews, but we should note it's a latex export

```suggestion
    doc.set_identifier(f"calkit-latex-export:{export_id}:{rev or ''}:{source}")
```

calkit/cli/latex.py
```diff
@@ -613,3 +613,426 @@ def _build_diff(
...
+    doc.save()
+    record = DocxExport(
```
petebachant: We may want to name this class more specifically, since it is for exporting a LaTeX-built PDF, not any old PDF.

calkit/docx.py
```diff
@@ -0,0 +1,679 @@
+"""Reading and writing Word documents for LaTeX review round trips."""
```
petebachant: This location implies it's for any docx handling, else it would be under `latex/docx.py`

```suggestion
"""Reading and writing Word documents in .docx format."""
```

calkit/docx.py
```diff
@@ -0,0 +1,679 @@
...
+from dataclasses import dataclass, field
+from xml.etree import ElementTree as ET
+
+W = "http://schemas.openxmlformats.org/wordprocessingml/2006/main"
+W14 = "http://schemas.microsoft.com/office/word/2010/wordml"
+W15 = "http://schemas.microsoft.com/office/word/2012/wordml"
+REL = "http://schemas.openxmlformats.org/officeDocument/2006/relationships"
+PKG_REL = "http://schemas.openxmlformats.org/package/2006/relationships"
+DC = "http://purl.org/dc/elements/1.1/"
+CK_NS = "https://calkit.org/review"
```
petebachant: These could use more descriptive names or comments

calkit/docx.py
```diff
@@ -0,0 +1,679 @@
...
+
+    def protection(self) -> str | None:
+        """The edit restriction in force, or None if none or unenforced."""
+        settings = ET.fromstring(self.parts["word/settings.xml"])
+        el = settings.find(_tag(W, "documentProtection"))
+        if el is None or el.get(_tag(W, "enforcement")) != "1":
+            return None
+        return el.get(_tag(W, "edit"))
+
+    def comments(self) -> list[Comment]:
```
petebachant, line 358: Wouldn't this be better as a property?

PR #1659, opened 2026-09-12
Push annotated tags along with commits

petebachant:
From Claude:

## Problem

`calkit push` runs a plain `git push`, which does not carry tags. `calkit save` delegates to `push`, so it has the same gap.

The effect is quiet: the push succeeds, `git status` is clean, and the local clone looks complete because the tags are right there. It only shows up for someone else — or on a fresh clone — where the history arrives without the labels.

That matters for a project that tags results. In the one I hit this on, experiment tags are how each row of a results table points at the commit that produced it, and several existed only on the machine that made them. I found it by checking `git ls-remote --tags` after a `calkit push` that reported success.

## Change

Add `--follow-tags` to the `git push` in `push()`.

`--follow-tags` is the conservative form: it sends **annotated** tags reachable from the refs being pushed. Unannotated scratch tags stay local, and so do tags on branches that are not being pushed — so this does not turn a push into a bulk tag publish.

It is skipped when the caller has already said something about tags via `--git-arg`, so `--git-arg --tags` and `--git-arg --no-follow-tags` both keep working.

## Test

`test_push_carries_annotated_tags` pushes to a real bare remote and asserts the annotated tag arrives and an unannotated one does not. It fails without the change and passes with it — I checked by stashing the fix and re-running.

### DIFF
--- a/calkit/cli/main/core.py
+++ b/calkit/cli/main/core.py
@@ -1672,17 +1672,13 @@ def push(
             git_cmd = ["git", "push"]
             if not no_recursive and "--recurse-submodules" not in git_args:
                 git_cmd.append("--recurse-submodules=on-demand")
-            # Annotated tags travel with the commits they mark. A plain
-            # `git push` leaves them behind, so a project that tags results
-            # ends up with tags that exist only on the machine that made
-            # them, and anyone cloning gets the history without the labels.
-            # `--follow-tags` is the conservative form: only annotated tags
-            # reachable from what is being pushed, so unannotated scratch
-            # tags and tags on other branches stay local.
-            if not any(
-                arg.startswith(("--follow-tags", "--no-follow-tags", "--tags"))
-                for arg in git_args
-            ):
+            # A plain `git push` leaves tags behind, so annotated tags end
+            # up only on the machine that made them
+            if not {
+                "--follow-tags",
+                "--no-follow-tags",
+                "--tags",
+            }.intersection(git_args):
                 git_cmd.append("--follow-tags")
             subprocess.check_call(git_cmd + git_args)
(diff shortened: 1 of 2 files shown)


calkit/cli/main/core.py
```diff
@@ -1672,6 +1672,18 @@ def push(
             git_cmd = ["git", "push"]
             if not no_recursive and "--recurse-submodules" not in git_args:
                 git_cmd.append("--recurse-submodules=on-demand")
+            # Annotated tags travel with the commits they mark. A plain
+            # `git push` leaves them behind, so a project that tags results
+            # ends up with tags that exist only on the machine that made
+            # them, and anyone cloning gets the history without the labels.
+            # `--follow-tags` is the conservative form: only annotated tags
+            # reachable from what is being pushed, so unannotated scratch
+            # tags and tags on other branches stay local.
```
petebachant: Shorten this comment

calkit/cli/main/core.py
```diff
@@ -1672,6 +1672,18 @@ def push(
...
+            if not any(
+                arg.startswith(("--follow-tags", "--no-follow-tags", "--tags"))
```
petebachant: Why `startswith` and not exact matching?

calkit/cli/main/core.py
petebachant: Shortened to two lines.

calkit/cli/main/core.py
petebachant: No good reason — exact matching is right and I've switched to it. These flags have no `=value` form, so `startswith` bought nothing and would have matched things like `--tagsomething`. Now a set intersection against `git_args`.

Issue #404, opened 2025-07-08
I don't want to have to run `nbstripout` manually on a repository


petebachant:

petebachant:
Executed notebooks in Git are now excluded automatically as of #1554, so we don't need to worry about that.

PR #1662, opened 2026-09-13
Make cleaned notebooks invariant to cell IDs

petebachant:
From Claude:

## Problem

A notebook stage depends on the cleaned copy of its notebook. `clean_notebook` strips outputs, execution counts and metadata, but kept cell `id`.

nbstripout rewrites cell IDs to ordinals and has **no smudge filter**. So for one commit there are two different byte streams:

- the working tree it was committed from: executed notebook, random hex IDs
- any fresh clone: the stripped blob, IDs `0`, `1`, `2`, …

Cleaning is deterministic, so it faithfully propagates that difference into the cleaned copy. The recorded dependency hash then only ever matches the machine that ran the stage, and the stage is stale on every clone — permanently, with `git status` clean and nothing reported as modified.

Measured on a project where this bit (86 KB executed source vs 39 KB stored):

| | md5 of cleaned copy |
|---|---|
| cleaned from the working tree | `62e8730f…` — matches `dvc.lock` |
| cleaned from what a clone gets | `78c8a121…` — does not |

The only differing field across all 16 cells was `id`.

This is the failure mode `ensure_path_is_not_filtered` already documents ("reports the stage stale forever"), reached through a dependency rather than a git-stored output, so that guard does not cover it.

## Fix

Drop cell IDs when cleaning. They are editor residue in the same sense as execution counts, and the cleaned copy exists only to be hashed — it is never executed.

## Note

Cleaned-notebook hashes change once, so notebook stages re-run the first time after upgrading.

The new test fails without the change and passes with it. `test_execute_notebook_julia` fails in my environment before and after, unrelated to this change.


petebachant:
Another potential solution for this could be to automatically manage users' `nbstripout` configs (#404)

This change will potentially invalidate notebook stages in the wild.

PR #1671, opened 2026-09-15
Make LaTeX diffs work with DVC-tracked inputs and stage build settings

petebachant:
From Claude: 

Pete and I worked through #1668, #1669, #1670, and #1672 together. They share a root cause: the built-in `diffs:` list on a `latex` stage couldn't build a correct diff for a paper whose figures and tables are DVC outputs, so the manual `git latexdiff` stage in MDOcean was the workaround, and each issue came out of that stage or the generated one. This PR makes the built-in diffs handle that case, so the manual stage and its helper scripts shouldn't be needed.

## Changes

**Each revision's DVC-tracked inputs (#1669)**

- `calkit latex diff` fetches each revision's DVC-tracked inputs as they were at that commit, from the cache or a remote, including a Calkit remote, and builds the marked-up document in the newer revision's checkout instead of beside the working copy. Before, both sides of a diff used today's figures and tables. A file in neither the cache nor a remote gives a warning, not an error.
- When a figure differs between revisions, the older side's `\includegraphics` is pointed at its own copy with a relative path, so it also resolves inside a container. A changed figure is now marked as changed, and `--graphics-markup=both` shows the old and new versions. Only paths relative to the document are handled, not `\graphicspath`.
- A latex stage passes its `inputs` to the diff command as `--input`, including outputs of other stages named with `from_stage_outputs`, which moving comparisons also depend on. Running `calkit latex diff` directly falls back to the inputs detected in the document plus any directories DVC tracks as a whole beside it.
- The check that skips a rebuild when nothing changed now covers the fetched files' hashes, the build options, and the rc file's contents. For a comparison against the working tree, it also covers the size and modification time of the working tree's input files.
- An output DVC records but doesn't store, e.g., MDOcean's glossary files that a `map-paths` stage copies in, has only its hash in `dvc.lock`, so no checkout can have it. It's now copied from the working tree when the content matches that hash. An input DVC can't list or fetch, e.g., because the remote needs a login, gives a warning with the underlying reason instead of a traceback.
- A verbatim input named by a macro parameter, e.g., `\verbatiminput{#1.wcsum}`, is now handled by editing the checked-out files directly. latexdiff's `--filter-script`, which calkit used before, reads the script's output back without decoding it, so it mangled anything outside Latin-1, e.g., the apostrophe in "Green’s". A comparison against the working tree still uses the filter, since those files can't be edited.

**Build settings and latexdiff options (#1669)**

[...]

Issue #1670, opened 2026-09-15
I want to see the actual error trace when a shell-command stage fails in docker env


rebeccamccabe:
Getting the actual error message (in my case the trace within `dev/latex/latexdiff-link-figs.sh`) would be more useful for debugging.
Currently it just shows that the command returned 1, and even a verbose run only gives the trace within calkit/dvc not within the command itself.
Stage definition:
```
make-aor-latexdiff:
      kind: shell-command
      environment: tex
      command: rm -rf git-latexdiff.* && DVC_REPO_ROOT="/work" DVC_CACHE_DIR="/work/.dvc/cache"
        DVC_DEPS_JSON="/work/aor-latexdiff-dvc-deps.json"
        git latexdiff --no-view --tmpdirprefix .
        --main pubs/applied-ocean-research-model/main.tex
        --whole-tree --ignore-latex-errors --latexmk
        --latexpand --makeatletter
        --graphics-markup=both
        --filter 'dev/latex/latexdiff-link-figs.sh'
        --prepare 'python3 ../new/dev/latex/latexdiff-prepare.py'
        -o pubs/applied-ocean-research-model/main-diff.pdf c767dee -- < /dev/null
```
Verbose error trace:
```
Running stage 'make-aor-latexdiff':                                                                                                            
> calkit xenv -n tex --no-check -- bash --noprofile --norc -c "rm -rf git-latexdiff.* && DVC_REPO_ROOT=\"/work\" DVC_CACHE_DIR=\"/work/.dvc/cache\" DVC_DEPS_JSON=\"/work/aor-latexdiff-dvc-deps.json\" git latexdiff --no-view --tmpdirprefix . --main pubs/applied-ocean-research-model/main.tex --whole-tree --ignore-latex-errors --latexmk --latexpand --makeatletter --graphics-markup=both --filter 'dev/latex/latexdiff-link-figs.sh' --prepare 'python3 ../new/dev/latex/latexdiff-prepare.py' -o pubs/applied-ocean-research-model/main-diff.pdf c767dee -- < /dev/null"
WARNING: DVC .dir object not found: pubs/applied-ocean-research-model/figs/manual/: 448281bb1c84a8e6fd960fd20398c33d
WARNING: DVC .dir object not found: pubs/applied-ocean-research-model/tables/: e69fd686d4c46b04d7d29066bf6b3fd0
Warning: Could not find file [tables/from-matlab/validation.tex]
[...]

Issue #1669, opened 2026-09-15
I want to inherit/control latexdiff options


rebeccamccabe:
The latexdiff command that ended up working for me (manually) was:
```
git latexdiff --no-view --tmpdirprefix .
        --main pubs/applied-ocean-research-model/main.tex
        --whole-tree --ignore-latex-errors --latexmk
        --latexpand --makeatletter
        --graphics-markup=both
        --filter 'dev/latex/latexdiff-link-figs.sh'
        --prepare 'python3 ../new/dev/latex/latexdiff-prepare.py'
        -o pubs/applied-ocean-research-model/main-diff.pdf
        c767dee -- < /dev/null
```

The filter and prepare options in mine can probably be omitted in the calkit latexdiff, since they are just hacky scripts I'm using to get the correct dvc versions between git revisions, and I assume calkit handles that better.

Notably, if a latexmk is already provided to the calkit stage, it should use `--latexmk` flag in the latexdiff, and the `--latexpand --makeatletter` and `
--graphics-markup=both` options are examples of general user-specified latexdiff options that may want to be passed, others could include the diff formatting etc.

PR #1675, opened 2026-09-15
Keep finding Calkit config and keyring secrets when a project changes HOME

petebachant:
From Claude: 

Pete and I found this while reproducing #1669 and #1670 on MDOcean for #1671. MDOcean sets `HOME: /tmp` in its `calkit.yaml` `env_vars`, and `calkit run` applies those to the process environment before running stages. Every Calkit command a stage runs then looked for its config under `/tmp`, and on macOS the keychain looked for the login keychain there too, so hub credentials disappeared. Any stage that needed the Calkit DVC remote or the hub failed with "No token found; Run 'calkit hub login' to authenticate", even though the user was logged in.

## Changes

- `set_env_vars` records the user's home directory in `CALKIT_USER_HOME` when a project's `env_vars` change it, e.g., through `HOME` or `USERPROFILE`. It doesn't replace a value an outer `calkit run` already set, and it isn't set when the home directory doesn't change.
- `calkit.config.get_user_home()` prefers `CALKIT_USER_HOME`. The config file path, the default hub lookup, and the environment check cache use it.
- Keyring calls run with `HOME` pointed at the user's home for their duration, then put back, since the macOS keychain reads `HOME` at the time of each call. Nothing outside `calkit.config` calls the keyring directly.
- The project's `HOME` still applies to everything else a stage runs, which is presumably why a project sets it.

## Testing

- New tests in `calkit/tests/test_config.py` and `calkit/tests/test_core.py` cover the config path following `CALKIT_USER_HOME`, a keyring call seeing the user's home with `HOME` restored afterwards, and `set_env_vars` recording the home only when it changes.
- Checked on MDOcean without any token override: after applying its `env_vars`, a child process, as a stage would be, found the hub token. Before this change, the same check failed with "No token found".
- The config tests pass locally, and there are no new ruff or mypy errors on changed lines.


petebachant:
@rebeccamccabe Claude dug this up looking at MDOcean's `HOME: /tmp` environmental variable setting. Not sure if that was causing you any issues. Maybe not if tokens are set via env var in GitHub Actions.

rebeccamccabe:
I never experienced this, even when running Calkit locally not in GitHub actions. Maybe it’s a Mac specific issue?

rebeccamccabe:
Actually, I just encountered it on CI:
https://github.com/symbiotic-engineering/MDOcean/actions/runs/35134695988/job/104924030159#step:18:1652

rebeccamccabe:
I tried using this PR in CI (merged it into the other branch I was using to create branch `rmccabe-subproject-login-utf8`) and it doesn't seem to fix the login issue:
https://github.com/symbiotic-engineering/MDOcean/actions/runs/35144986541/job/105020561316#step:18:16835

petebachant:
That looks to be a different issue, related to LaTeX diffs needing old versions of figures in DVC. Changes to the run Action are on this branch. You may want to update the workflow to `uses: calkit/calkit/actions/run@main` after merging.

PR #1695, opened 2026-09-18
Improve quickstart

petebachant:
The current quickstart isn't really that quick, and the `xr` commands are unique to a given project. That probably belongs elsewhere. We want something quick that gets a user up and running:

1. New project
2. Import dataset
3. Ask research question
4. Explore data?
5. Analyze/plot (perhaps ask a coding agent to do it)
6. Compile paper ready for iteration until submission (without giant system installs or Docker pulls)

Should have a path towards expanding into an entire PhD's worth of research, or is this just one study that gets lumped into a PhD later?

Also trying out using ADRs on this branch. Deleted because we didn't make important architectural changes--scaled these back.

## TODO

- [ ] Rewrite prose by hand.


images/latex/README.md
```diff
@@ -0,0 +1,80 @@
+# LaTeX image
+
+`ghcr.io/calkit/latex`, the image a LaTeX stage falls back to when the
+machine has no TeX of its own. TinyTeX plus a curated set of packages and
+`latexdiff`, which comes to about 766 MB against roughly 9 GB for
+`texlive/texlive:latest-full`.
+
+Originally developed at
+[`calkit/tinytex-latexmk-docker`](https://github.com/calkit/tinytex-latexmk-docker),
+and based on [carteakey/tinytex-docker](https://github.com/carteakey/tinytex-docker).
```
petebachant, line 17: We don't need the original Calkit context, only the reference to carteakey's repo.

images/latex/README.md
```diff
@@ -0,0 +1,80 @@
...
+script, and the `perl` in the first layer is what makes it work here, on a
+machine that may have no Perl of its own.
+
+## Packages a document needs beyond this set
+
+Install them into the mounted cache rather than rebuilding the image:
+
+```sh
+tlmgr --usermode install <package>
+```
```
petebachant: Users who are only using the Docker image won't be able to do this unless the run it in the image, perhaps with `calkit xenv`.

images/latex/README.md
```diff
@@ -0,0 +1,80 @@
...
+
+A project says what it needs in `calkit.yaml`, and Calkit installs them:
+
+```yaml
+environments:
+  tex:
+    kind: tinytex
```
petebachant: This is not a system tinytex install, which could be an option someday. We should be clear.

```suggestion
    kind: tinytex-docker
```

images/latex/README.md
```diff
@@ -0,0 +1,80 @@
...
+    packages:
+      - fancyhdr
+```
+
+## Running it directly
```
petebachant, line 132: We could add the `calkit xenv` equivalent here.

images/latex/README.md
```diff
@@ -0,0 +1,80 @@
...
+starts empty and would hide whatever the image had put there.
+
+## Releasing
+
+Push a tag and the workflow builds and pushes it:
+
+```sh
+git tag latex-image/v1.0.0
+git push origin latex-image/v1.0.0
+```
```
petebachant: This should be a GitHub release like the rest, not a tag push from a dev's machine.

calkit/docker.py
```diff
@@ -831,6 +937,7 @@ def build_lock(
 
     Key order is fixed so that a lock written for a platform from a registry
     matches byte-for-byte the one that platform would write for itself.
+
```
petebachant:
```suggestion
```

docs/index.md
```diff
@@ -133,6 +133,14 @@ which is a critical feature to have when using generative AI.
   Simply call `calkit run`.
   Compose your pipeline from many different kinds of stages,
   including simple scripts, commands, Jupyter Notebooks, LaTeX, and more.
+- Tools for automatically building reproducible pipelines from work you're
+  already doing.
+  Run a script, notebook, or LaTeX document with
+  [`calkit xr`](pipeline/index.md#automatic-stage-and-environment-detection)
+  and it becomes a pipeline stage,
+  with its environment, inputs, and outputs detected and recorded,
+  so an existing project becomes reproducible a command at a time rather
```
petebachant:
```suggestion
  so an existing project becomes reproducible one command at a time rather
```

docs/quickstart.md
```diff
@@ -5,103 +5,151 @@
...
-as reproducible pipeline stages.
-For example:
+## Create a project
 
 ```sh
-calkit xr scripts/analyze.py
-
-calkit xr notebooks/plot.ipynb
+calkit new project my-research \
+    --title "My research" \
```
petebachant: Project title can be changed later, but name is more difficult, so be general. It's also okay to consider the project a big container for multiple investigations, papers, a thesis. In this case, we use a simple `phd` directory.

```suggestion
calkit new project phd \
    --title "PhD research" \
```

docs/quickstart.md
```diff
@@ -5,103 +5,151 @@
...
+
+Add `--hub` to also create it on [a Calkit hub](hub/index.md), which is
+how projects get backed up and shared.
+It needs an account, and can be set up later, so leave it off for now if
+you just want to see the thing work.
+
+## Look at the question
+
+Open `calkit.yaml`.
+Near the top is what this project is for:
```
petebachant, line 43: Should we have the users run `calkit list questions` and/or `calkit check questions`? The latter would show that we have a question with stale evidence. We should probably have them run `calkit status` to see what outputs are stale. I believe a new project will start with some staleness.

Maybe this is a good argument for putting question status into `status`, but more of a summary like:

```
5 questions, 2 unanswered, 1 with stale evidence, 2 with no evidence, 1 with non-parameterized values.
```

docs/quickstart.md
petebachant: From Claude: took the suggestion -- the project is `phd` with title "PhD research" now, and the section says the title is easy to change later but the name isn't, and that one project can hold several investigations (a question and a paper per study).

docs/quickstart.md
petebachant: From Claude: added both. `calkit list questions` is now in "Look at the question", showing the answer with $R^2$ read out of `results/summary.json` rather than typed in, and "Change something" now runs `calkit check questions` after the edit, which reports the answer as stale because the stage behind its evidence is.

I also took the summary idea: `calkit status` has a Questions section now.

```
--------------------------- Questions ----------------------------
1 question, 1 with stale evidence
Run 'calkit check questions' for detail.
```

It reads `N questions` followed by only the counts that aren't zero (unanswered, with stale evidence, with missing evidence, with broken references, resting on a frozen stage, with no evidence), or `all answered with current evidence ✅` when there's nothing to say. It reuses the stale stages already computed for the pipeline section rather than asking DVC twice, and the section is skipped entirely when a project has no questions.

Two related changes fell out of it: `calkit new project --template` was dropping the template's questions, which left a new project with no question to look at, so it keeps them now; and `calkit list questions` was printing `explanation: None` / `notes: None` for every field a question doesn't use.

calkit/cli/main/core.py
```diff
@@ -1106,6 +1151,37 @@ def add(
...
+                    typer.echo(
+                        f"Ignoring {path} since it's a DVC pipeline output"
+                    )
+                    # Where and how DVC itself would have written the entry
+                    if dvc_scm is None:
+                        dvc_scm = calkit.dvc.get_dvc_repo().scm
+                    gitignore = dvc_scm.ignore(os.path.abspath(locked_out))
+                    if gitignore:
+                        subprocess.call(["git", "add", gitignore])
+                    subprocess.call(["git", "add", "dvc.lock"])
```
petebachant: Can use `git.Repo` for these commands

calkit/cli/main/core.py
```diff
@@ -3434,6 +3566,12 @@ def save_env_check_cache() -> None:
                 if isinstance(value, str):
                     value = os.path.expandvars(value)
                 docker_cmd += ["-e", f"{key}={value}"]
+        # Set by whatever is calling xenv, e.g., 'calkit latex build'
+        # stamping a reproducible date into the PDF, and pointless unless
+        # it reaches the container
+        for key in ["SOURCE_DATE_EPOCH", "FORCE_SOURCE_DATE"]:
+            if key in os.environ and key not in env_vars:
+                docker_cmd += ["-e", f"{key}={os.environ[key]}"]
```
petebachant: This is kind of weird. We have a list of protected env vars that we'll pass to Docker envs if set? Maybe this should actually be an `--env-var SOURCE_DATA_EPOCH=whatever` option in `xenv`, so we can set them in the environment in which we execute. Then `calkit latex build` can use that in `xenv`. This option can also set them in the other subprocesses we call for other env kinds.

calkit/cli/config.py
```diff
@@ -189,9 +189,17 @@ def setup_remote(
...
     """Set up the Calkit hub as the default DVC remote and store a token
     in the local config.
+
+    Deprecated: this configures the project, not Calkit itself, which is
+    what the rest of this app is for.
     """
     from git.exc import InvalidGitRepositoryError
 
+    warn(
+        "'calkit config remote' is deprecated; use 'calkit update hub', "
```
petebachant, line 199: Maybe `calkit update remote` also makes sense as a command. I also kind of like `calkit update project hub`, but we are going with the entity in update applying to the project primarily. This could make things like `calkit update <app>` irrational, since they are updating the system, not the project. `calkit update project name` reads a little nicer than `calkit update name`, but `update_project` would be another nested CLI layer. It brings us closer to sounding like English, but more verbose.

calkit/cli/latex.py
```diff
@@ -137,12 +137,27 @@ def tokens_from_format_string(fmt: str):
             json2latex.dump(cmd_name, formatted, f)
 
 
+def _tex_env(source_date_epoch: str | None) -> dict[str, str] | None:
```
petebachant: Environment and environmental variables could be confused.

```suggestion
def _tex_env_vars(source_date_epoch: str | None) -> dict[str, str] | None:
```

calkit/cli/new.py
```diff
@@ -1571,10 +1573,12 @@ def new_publication(
...
     # Create environment if applicable
     if env_name is not None and template_type == "latex":
+        from calkit.latex import DEFAULT_LATEX_IMAGE
+
         env = dict(
             kind="docker",
-            image="texlive/texlive:latest-full",
-            description="TeXlive full.",
+            image=DEFAULT_LATEX_IMAGE,
+            description="TeX Live via Calkit's LaTeX image.",
```
petebachant: We could use a `calkit.latex.DEFAULT_CALKIT_ENVIRONMENT` to encapsulate this whole dict for use elsewhere.

petebachant: It is duplicated a bit.

calkit/cli/main/core.py
petebachant: From Claude: done, both use the repo already in scope now (`repo.git.add(gitignore)` and `repo.git.add("dvc.lock")`). The other `git add` calls in that function predate this and still shell out; say the word and I'll convert them too.

calkit/cli/main/core.py
petebachant: From Claude: agreed, that was a list of blessed names, which is a rule nobody can see from the call site. Replaced with a general `--env-var KEY=VALUE` option on `xenv`, repeatable.

It's set in the process the command runs in, so every environment kind's subprocess inherits it, and handed to a container with `-e` for the kinds that run in one, since a container inherits nothing from here. Malformed input is refused with a message naming the form.

`calkit latex build` now passes `--env-var SOURCE_DATE_EPOCH=… --env-var FORCE_SOURCE_DATE=1` through `xenv`, and the same pair as `-e` when it runs the image directly, so the three paths say the same thing in the way each one needs. `_tex_env_vars` is now just those variables, which the system path merges into the environment it passes to the subprocess.

Tested for a system and a Docker environment, with and without the option, plus the malformed case, and the LaTeX fetching still works end to end through a Docker environment.

calkit/cli/latex.py
petebachant: From Claude: renamed to `_tex_env_vars`, and its docstring now says environmental variables rather than environment.

calkit/cli/new.py
petebachant: From Claude: added `calkit.latex.DEFAULT_LATEX_ENVIRONMENT`, the whole dict, and used it here, in the Overleaf import, in environment detection, and in the hub's Overleaf import. Each copies it, since what gets written into a project is the caller's to amend.

petebachant: From Claude: it was, in four places. All of them use `DEFAULT_LATEX_ENVIRONMENT` now.

calkit/cli/config.py
petebachant: From Claude: `calkit update remote` already exists as an alias for `hub` (`calkit/cli/update.py:1607`), so that spelling works today.

On `update project hub`: no change made, since this is your call on the CLI's shape. The tension I see from having just used it is that `update` already holds both kinds of object — `calkit update devcontainer` and `calkit update vscode-config` write project files, while `calkit update <app>` updates the machine. A `project` layer would name the first group honestly and leave the second reading oddly as a sibling, so it's worth deciding on the grouping rather than on how any single command reads. Happy to do the rename if you want it.

calkit/cli/main/core.py
petebachant: From Claude: converted the rest too. All the staging in `add` now goes through `repo.git.add`, and the commit through `repo.git.commit`.

Two things shelling out was hiding, both kept working rather than changed:

- `subprocess.call` ignored a non-zero exit, so a path Git refuses, e.g., one a .gitignore excludes, passed silently. GitPython raises instead, so staging goes through a small helper that catches it and warns with what Git said, and the run carries on. A commit with nothing staged is a normal outcome and stays quiet.
- GitPython runs Git from the repo's root, while these paths are relative to the working directory, so `calkit save deep.txt` from a subdirectory staged nothing. `test_commit` caught it; paths are resolved before they're handed over.

Tests: all 245 in `calkit/tests/cli` pass, and a new one covers the refused-path case.

PR #1704, opened 2026-09-21
Add TinyTeX image for building LaTeX documents

### LINKED ISSUE #407: I want a smaller LaTeX Docker image, or just a lighter weight LaTeX compilation process, perhaps in the cloud, TinyTeX, tectonic
https://gist.github.com/wkrea/b91e3d14f35d741cf6b05e57dfad8faf

Could get all of the functionality with less bloat.

Tectonic and TinyTeX are other alternatives. Tectonic is nice because it automatically fetches packages, and we could mount a local folder in the project for that, so they persist between runs.

We could make environments optional for LaTeX stages, and if not supplied, we dynamically pick the right way to run LaTeX: Tectonic, Docker, or system latexmk.

WASM: https://github.com/swimmingbrain/texbrain

petebachant:
From Claude:

Adds `images/latex`, published as `ghcr.io/calkit/latex`, so a LaTeX stage doesn't need `texlive/texlive:latest-full`.

**803 MB against roughly 9 GB.** TinyTeX plus a curated set of packages rather than whole TeX Live collections, which would have made it 5.29 GB and barely better than what it replaces.

## What it carries

Papers in these classes build with nothing installed at run time: Elsevier (`elsarticle`), APS (`revtex4-2`), IEEE (`IEEEtran`), ACM (`acmart`), ACS (`achemso`), MNRAS, Springer LNCS (`llncs`), AMS (`amsart`), plus the JFM template in this repo and the AASTeX class that ships with a paper rather than through TeX Live.

The package list was derived by compiling a document in each class against a bare TinyTeX and installing whatever it asked for. Two of them cannot be fetched on demand at all, which is why fetch-on-demand isn't enough on its own:

- `aastex631.cls` refuses to load without `revtex4-1` and stops without reporting a missing file
- `newtx` needs `txfonts`' metrics, which surface as `Metric (TFM) file not found` rather than as a missing `.sty`

`latexdiff` is included, with the Perl it needs, so `calkit latex diff` works in the container.

## Two bugs this fixes

The earlier standalone image didn't work, for two reasons:

- Its documented invocation mounted a cache over `/root/.TinyTeX`, where the distribution lives, leaving no `latexmk` or `pdflatex` at all. The cache now mounts at `/texmf`, which is `TEXMFHOME`.
- TinyTeX under `/root` (mode 0700) was unreadable to a non-root user, so running as the invoking user failed outright and running as root left PDFs owned by root. It now installs to `/opt` with `a+rX`.

The user tree is initialized by `entrypoint.sh` at run time rather than in a build layer, since a mounted cache starts empty and would hide anything the image had put there.

## Testing

On every pull request touching `images/latex`: tool checks as root and as a non-root user, a document built with the cache mounted, an ownership check on the PDF, a package installed into the mounted cache, a paper built in each of the eight journal classes, and the JFM template.

The journal builds are the guard that matters. The package list is empirical, so without them, trimming a package only one class needs would go unnoticed until someone submitted a paper.

Local builds were arm64; the first CI run is the first amd64 build.

## Releasing

Tag `latex-image/v1.0.0` to build multi-platform and push to ghcr.io with a provenance attestation. Nothing in Calkit points at the image yet; that switch is one constant, in the quickstart branch.

Resolves #407


.github/workflows/test-latex-image.yml
petebachant, line 1: Testing the latex image and building/pushing the latex image should probably be two different workflows like with the others, where published releases trigger the publishing of the image.

petebachant: From Claude: Split into `test-latex-image.yml` (pull_request + workflow_dispatch) and `publish-latex-image.yml` (release: types: [published], guarded on `latex-image/v`), matching `test-assistant.yml` and `publish-vscode-ext.yml`. Publishing is now a published release rather than a bare tag push.

.github/workflows/test-latex-image.yml
```diff
@@ -0,0 +1,225 @@
...
+            set -eux
+            tlmgr --usermode install lipsum
+            kpsewhich lipsum.sty
+          '
+          test -n "$(find "$cache" -name lipsum.sty -print -quit)"
+      # Every journal class the image claims to carry has to build with
+      # nothing installed at run time. Trimming a package that only one
+      # class needs would otherwise go unnoticed until someone submitted
+      # a paper.
+      - name: Build a paper in each journal class
```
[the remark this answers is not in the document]
petebachant: From Claude: Correct, and rather than add a fixture I removed the claim, because it can't honestly be tested here: AASTeX's class isn't in TeX Live and isn't redistributable into this repo. The README now says the support is real but unguarded, and that removing `revtex4-1` would break AASTeX without failing a test. Covering it properly needs the downloading-template mechanism we're planning for JFM, which is separate work.

images/latex/Dockerfile
```diff
@@ -0,0 +1,100 @@
...
+# to be writable by whoever runs the container, not just root.
+ENV TEXMFHOME=/texmf
+RUN mkdir -p /texmf && chmod 1777 /texmf
+
+RUN fmtutil-sys --all
+
+# A curated core rather than whole collections: the tools, the classes
+# for journals people actually submit to, and what those classes need. Determined by
+# compiling an AASTeX 6.3.1 paper and the JFM template against the
+# minimal image and installing whatever each asked for, one at a time.
```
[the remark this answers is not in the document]
petebachant: From Claude: Fixed, along with two more over-80 lines the review didn't flag: the `ENV PATH` line (now split across two `ENV`s) and a condition in `entrypoint.sh`.

images/latex/README.md
```diff
@@ -0,0 +1,93 @@
+# LaTeX image
+
+`ghcr.io/calkit/latex`, the image a LaTeX stage falls back to when the
+machine has no TeX of its own. TinyTeX plus a curated set of packages and
+`latexdiff`, which comes to about 766 MB against roughly 9 GB for
+`texlive/texlive:latest-full`.
```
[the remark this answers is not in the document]
petebachant: From Claude: Right on both counts. The README now opens by saying nothing in Calkit points at the image yet and that wiring it up is a separate change, and the stale `DEFAULT_LATEX_IMAGE` paragraph is gone.

images/latex/README.md
```diff
@@ -0,0 +1,93 @@
...
+```
+
+A project says what it needs in `calkit.yaml`, and Calkit installs them:
+
+```yaml
+environments:
+  tex:
+    kind: tinytex-docker
+    packages:
+      - fancyhdr
```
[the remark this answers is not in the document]
petebachant: From Claude: Correct, `tinytex-docker` only exists on the quickstart branch, so that example would have failed validation. Replaced with just the `tlmgr --usermode install` path and a note that the package lands in whatever is mounted at `TEXMFHOME`.

petebachant:
From Claude: Pushed the review round. Beyond the comments above, testing the image as a non-root user turned up a bug that would have shipped: Elsevier and MNRAS ask for EC fonts, which this install carried only as bitmaps, so TeX shelled out to `mktexpk` at build time and wrote into `/opt/.TinyTeX/texmf-var` — not writable by the invoking user.

```
mkdir: cannot create directory '/opt/.TinyTeX/texmf-var/fonts/pk/ljfour': Permission denied
!pdfTeX error: pdflatex (file ecbx1000): Font ecbx1000 at 600 not found
```

Every earlier sweep of mine ran as root, which hid it. Notably CI passed `elsarticle` on amd64 while it failed locally on arm64, so the behaviour was environment-dependent rather than reliably caught.

Fixed two ways: `cm-super` supplies Type 1 EC fonts so nothing is generated at run time (and the PDFs are vector rather than bitmap), and `texmf-var` is writable for anything else that asks. All eight journal classes now build as a non-root user. The image goes from 803 MB to 991 MB, still about a ninth of `texlive/texlive:latest-full`.

The earlier CI failure was my `acmart` fixture, not the image — recent `acmart` requires `\country{}` in an affiliation, and CI's fresh `tlmgr install` had a newer version than my local build. That same drift is now noted in the README as the reason to pin a tag rather than track `:latest`.

.github/workflows/publish-latex-image.yml
```diff
@@ -0,0 +1,65 @@
+name: Publish the LaTeX image
```
petebachant:
```suggestion
name: Publish LaTeX image
```

.github/workflows/test-latex-image.yml
```diff
@@ -0,0 +1,173 @@
+name: Test the LaTeX image
```
petebachant:
```suggestion
name: Test LaTeX image
```

.github/workflows/publish-latex-image.yml
petebachant: From Claude: Applied — matches `Test assistant` and `Publish VS Code extension`.

.github/workflows/test-latex-image.yml
petebachant: From Claude: Applied — matches `Test assistant` and `Publish VS Code extension`.

PR #1706, opened 2026-09-21
Deploy staging only when the hub changes

petebachant:
From Claude:

Merging the LaTeX image PR deployed the staging hub, which it had no reason to do. The workflow fires on any push to `main` with no path filter, so every merge redeploys staging regardless of what changed.

The filter has to account for how the hub is actually built. `hub/docker-compose.yml` builds the backend with `context: ..`, the repository root, and the Dockerfile copies in more than `hub/`:

```dockerfile
COPY ./pyproject.toml ./uv.lock ./README.md /app/
COPY ./calkit /app/calkit
COPY ./agent-plugin /app/agent-plugin
COPY ./hub/backend /app/hub/backend
```

So filtering on `hub/**` alone would have skipped deploys that genuinely change the hub image. The filter covers each of those paths plus the workflow itself.

Production is unaffected: it deploys on a published release rather than on push.


.github/workflows/deploy-hub-staging.yml
```diff
@@ -12,6 +12,19 @@ on:
...
+    # package and the agent plugin as well as hub/ itself. Without this,
+    # anything merged to main deployed staging, e.g., a change to a
+    # Docker image we publish.
+    paths:
+      - "hub/**"
+      - "calkit/**"
+      - "agent-plugin/**"
+      - "pyproject.toml"
+      - "uv.lock"
+      - "README.md"
```
petebachant, line 26: Why the README?

.github/workflows/deploy-hub-staging.yml
```diff
@@ -12,6 +12,19 @@ on:
...
       - main
+    # Only what the hub is actually built from. The backend image builds
+    # with the repository root as its context, so it takes in the Calkit
+    # package and the agent plugin as well as hub/ itself. Without this,
+    # anything merged to main deployed staging, e.g., a change to a
+    # Docker image we publish.
+    paths:
+      - "hub/**"
+      - "calkit/**"
+      - "agent-plugin/**"
```
petebachant, line 23: Why this path?

petebachant: From Claude: `agent-plugin` goes into the backend image. `hub/backend/Dockerfile` builds with the repository root as its context and copies it in:

```dockerfile
COPY ./pyproject.toml ./uv.lock ./README.md /app/
COPY ./calkit /app/calkit
COPY ./agent-plugin /app/agent-plugin
COPY ./hub/backend /app/hub/backend
```

So a change there changes what runs on staging, and leaving it out would mean an agent-plugin change silently not deploying.

.github/workflows/deploy-hub-staging.yml
petebachant: From Claude: Same reason mechanically, but weaker, and I'd drop it if you prefer.

`README.md` is copied into the image because `pyproject.toml` declares `readme = "README.md"`, so the build fails without it and the built image differs when it changes. But nothing about the running hub changes, so deploying staging for a README edit is a rebuild with no observable effect.

Dropping it means a README-only commit doesn't deploy, which seems like what you'd want. Say the word and I'll remove it.

PR #1717, opened 2026-09-22
Resolve value and document evidence on the hub, and let questions be deleted

petebachant:
From Claude:

A project migrated as `calkit check questions` asks (`result` + `key` → `value`, write-ups cited as `publication` evidence) passes that check 12/12 yet shows every question with **Missing** evidence on the hub. Three hub bugs, all confirmed against the live API for `petebachant/clima-gpu-profiling`:

1. **`value` evidence is dropped.** `_build_question_evidence` accepts only `figure`, `result`, `table` and `publication` and skips the rest, so none of that project's 31 value entries come back. (Templates still render, since `_evidence_values` uses `is_value_evidence`.)
2. **Undeclared publications are always Missing.** They resolve only against top-level `publications:` at the evidence's ref. The CLI checks by path, and the `Question` model says evidence needn't be declared first.
3. **Saving a question strips fields.** `_apply_question_update` and the editor rebuild each entry from `kind`, `path`, `key` (results only), `explanation` and `git_ref`, so an edit drops a value's `name`, which breaks every template that uses it, and a publication's `section`/`label`. `value` would also fail the `kind` Literal with a 422.

## Changes

- **`document` evidence kind** (CLI model, checker, schema): a write-up cited by path with an optional `section`. The checker treats it like a publication: it has to exist, but needn't be produced by a stage.
- **Hub resolution:** `value` resolves like a keyed result. `document`, and a `publication` nobody declared, are missing only if the path is absent from the tree (or a `.dvc` pointer / `dvc.lock` out) at the evidence's ref. They open in the file viewer and render inline in the question's detail pane.
- **Round-trip:** `QuestionEvidence`/`QuestionEvidencePost` gain `name`, `section`, `label`; the editor carries them in hidden fields, as it already does `git_ref`.
- **Deleting a question:** `DELETE /projects/{owner}/{project}/questions/{number}` removes the entry (dropping `questions:` when it empties), commits, pushes and reindexes; the index already closes up the numbering. The editor gets a Delete button using `DeleteAlert`, and the question page returns to the project afterwards.

## Testing

Written on a RHEL 8 host without Docker, so:

- CLI: `test_questions`, `models/test_core`, `test_schema`, `cli/test_list` pass. `cli/test_check` has 7 Docker-environment failures that are identical on `main`.
- Hub backend: the new evidence and save tests plus the existing evidence tests pass with `--noconftest` and dummy settings. **`test_delete_project_question` needs the test database and has not been run**; CI's `make test-backend` is its first run.
- Frontend: `tsc --noEmit` clean. The client was regenerated with the repo's own `openapi-ts` config from a spec dumped by `app.main.app.openapi()` instead of the served one. The regenerated client matched `main` exactly apart from these changes, so it should pass the stale-client check.
- Biome's glibc build won't run on this host, so formatting used the musl build of the same version (1.8.3), and the commit skipped only the `biome-format-hub-frontend` hook.

## Left for a human

The evidence-kinds list in `docs/questions.md` (line 41) needs a bullet for `document`.

https://claude.ai/code/session_01FYv7qDRgpuycyDQtmL96GH


calkit/questions.py
```diff
@@ -955,6 +955,9 @@ def check_evidence(
     if kind == "publication":
         out.status, out.message = _check_publication_label(ev, ck_info, wdir)
         return out
+    # Written by hand, so there is no stage to attribute it to
```
petebachant: Documents should potentially be built by stages. They could even be Markdown stages themselves, allowing them to be more self-contained calculations.

petebachant: From Claude:

Agreed. Documents now go through the same stage checks as other evidence in 3a64fcea: one built by a stage is stale or frozen with it, and that includes a Markdown stage, which names the document as its `target_path` rather than an output and runs as sub-stages like `doc/analyze`. A hand-written document still only has to exist, so it isn't reported as unattributed. One gap remains on the hub: it finds a stage only through `dvc.lock` outs, so it won't yet mark a Markdown-built document stale, since that needs the pipeline config at each ref and the evidence lookups don't carry it.

hub/frontend/src/components/Projects/QuestionModal.tsx
```diff
@@ -387,6 +414,15 @@ function EvidenceDetail({
...
     return <TableView table={table} maxHeight="calc(92vh - 300px)" />
   }
+  if (citedByPath) {
+    if (documentRequest.isPending && evidence.stale_reason !== "missing") {
+      return <LoadingSpinner height="200px" />
+    }
+    if (!documentRequest.data) {
+      return <NotFound evidence={evidence} />
+    }
+    return <FileContent item={documentRequest.data} />
```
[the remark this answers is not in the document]
petebachant: From Claude:

Fixed in 3a64fcea: when the contents come back with only a URL, the modal fetches the text and hands `FileContent` the encoded content, so Markdown and plain text render as before. A failed fetch says so and points at the file link instead of rendering blank.

calkit/models/core.py
```diff
@@ -1817,6 +1817,30 @@ class PublicationEvidence(BaseModel):
     )
 
 
+class DocumentEvidence(BaseModel):
+    """Evidence in the form of a document cited by path, e.g., a Markdown
+    write-up, without declaring it as a publication.
+    """
+
+    kind: Literal["document"] = "document"
```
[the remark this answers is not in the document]
petebachant: From Claude:

Added in fd99f693: a bullet in the kinds list, and a short example under 'Pointing at the publication' for reasoning kept in a document rather than a publication.

petebachant:
From Claude: the two overview items without inline comments are also in 3a64fcea. Keyless `value` evidence is now marked missing, since it names no value. And Delete is disabled while browsing a ref other than the default branch, because deletion writes to the default branch and question numbers there may not match the ref on screen. (Chakra's `Button` defaults to `type="button"`, so Delete doesn't submit the edit form.)

calkit/questions.py
petebachant: From Claude:

Went further in e10529ef after your follow-up: a document no stage builds is now an **error**, the same as a value no stage computes, unless it's declared with `imported_from`. A Markdown stage that injects its numbers is the usual way to build one, and the docs and `DocumentEvidence` docstring say so. For Quarto (10a28c12): a Quarto stage lists its rendered HTML/PDF under `outputs`, so citing that resolves to the stage and goes stale with it. Citing the `.qmd` is an error whose message names the stage and suggests its first output, since the source's numbers only exist once rendered.

